PR #12, opened 2019-08-19
Consume toolkit from npmjs

damccorm:


.gitignore
```diff
@@ -1,2 +1,2 @@
-!node_modules/
-__tests__/runner/*
+!node_modules/
```
stephenmichaelf: Is there a line endings issue?

damccorm: Yes, but I think it might actually be the fault of the repo not this PR. I ran `git add * --renormalize` on all of these PRs, this is the only one that had issues. [Renormalize](https://git-scm.com/docs/git-add#Documentation/git-add.txt---renormalize) just readds the files with the configuration you have on your machine. In my case, I have `core.autocrlf` set to true, which means it will get added with `LF` endings (which is what we want)

Issue #11, opened 2019-08-19
Use recommended release strategy


damccorm:
We should be using the release strategy specified by https://github.com/actions/toolkit/blob/master/docs/action-versioning.md#recommendations

To get there, we need to:
1) Create a `releases/v1` branch and move the `v1` tag and future releases there
2) Warn that we will deprecate the master branch and that you should attach to `v1` instead. We will support master in a warned but deprecated state for ~1 month.
3) Switch the warning to an error on master. Remove `node_modules` from the master branch. We will stay in this state for ~1 month.
4) Remove the error on master.

During steps 2 and 3, we need to be extra careful to not check the warning/error into the releases branch.

bryanmacfarlane:
We're not doing this anymore.  We're moving toward ncc.  See javascript-action PRs from me and talk to Josh

bryanmacfarlane:
At that point it's just code and branches and the branching is up to you.

konradpabjan:
A releases branch, `release/v1` has been created to track of a stable release version. The `v1` tag has also been adjusted to now point to that branch. In the future if there are big changes then `releases/v2` etc. will be created along with new tags. The master branch will remain functional with the latest and greatest changes changes. After changes in the master branch are validated and confirmed to be working they will be ported over to the releases branch.

`Node_modules` have been removed from the master and release branch and we now rely on ncc to create a single js file for release. Since the master branch will continue to function, there is no need for deprecation warnings or error messages. We document that using a reference to the master branch could cause breaking changes in the future so no further action is needed: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsuses

PR #35, opened 2019-11-05
Remove node modules and configure ncc

konradpabjan:
Relevant Issue: https://github.com/actions/setup-python/issues/11

Removing `node_modules/` and `lib/` from the master branch. In the releases branch, `ncc` is now being used to compile all of the necessary files into a single file `dist/index.js`

Going off of examples here: https://github.com/actions/cache/blob/releases/v1/package.json and here: https://github.com/actions/javascript-action/pull/5

Tested in private repository with existing action: `uses: actions/setup-python@v1` and with the new changes `uses: actions/setup-python@konradpabjan/ncc`, and confirming that it is working as expected:

Full yaml:

```
name: Python
on: [push]
jobs:
  build:
    runs-on: ubuntu-latest  
    strategy:
      matrix:
        version: [ '2.x', '3.x', 'pypy2', 'pypy3' ]
    name: Python ${{ matrix.python-version }} sample
    steps:
      - uses: actions/checkout@master
      - name: Setup python
        uses: actions/setup-python@konradpabjan/ncc
        with:
          version: ${{ matrix.python-version }}
          architecture: x64
      - run: python my_script.py
```

### Passing runs

image


konradpabjan:
A lot of changes here since I ended up deleting `node_modules` and `lib`. Main files to look out for:
- action.yml
- .gitignore
- package.json

joshmgross:
With early documentation, we often recommended using actions at master (`setup-python@master`), and this will be a breaking change for them (action will just stop working).

Do we know how many users are using this action on the master branch? Are we prepared/willing to break all of those users?

If not, we can just put `dist/index.js` in master as well

action.yml
```diff
@@ -10,4 +10,7 @@ inputs:
     default: 'x64'
 runs:
   using: 'node12'
-  main: 'lib/setup-python.js'
+  main: 'dist/index.js'
+branding:
+  icon: 'play'
```
joshmgross: We're not super consistent about which icons we use for the setup actions, I personally would prefer `code`, see https://help.github.com/en/github/automating-your-workflow-with-github-actions/metadata-syntax-for-github-actions#branding

package.json
```diff
@@ -41,8 +43,7 @@
   "husky": {
     "skipCI": true,
     "hooks": {
-      "pre-commit": "npm run build && npm run format",
-      "post-commit": "npm prune --production && git add node_modules/* && git commit -m \"Husky commit correct node modules\""
+      "pre-commit": "npm run build && npm run format && npm run format-check"
```
joshmgross: `format-check` isn't required here. `format` will handle everything

brcrista:
@joshmgross to your earlier comment, from #11:
> We will support master in a warned but deprecated state for ~1 month.

@konradpabjan have we deprecated master yet?

konradpabjan:
> If not, we can just put `dist/index.js` in master as well

The plan all along was to put `dist/index.js` into master as well as the releases branch so there should be no deprecation or breaking changes. The releases branch will have long term supported stable changes, while master will have the latest and greatest that can sometimes break (the changes in this PR don't break anything). After some period of time in the master branch, changes can be ported over to the releases branch.

What's described at the top of #11, is no longer 100% accurate. Originally `node_modules` was going to be removed in the master branch and kept inside the releases branch so that the master branch wasn't going to be usable. This was all before `ncc` was even suggested and we've decided against using this approach.

action.yml
konradpabjan: I noticed the branding was blank so I decided to fill it out. I opted for yellow because python is yellow/blue

I like the `code` icon so I've put that one in. We currently use a green play icon so that's why I decided to just use that one. But code seems better: https://github.com/marketplace/actions/setup-python-environment

konradpabjan:
> @konradpabjan have we deprecated master yet?

So we're not going to be deprecating master. I makes it possible for people to use it and test it out early if they really want. It also just seems weird to always have a not working action in the master branch.

What we could do is add a warning message inside `setup-python.ts` just inside the master branch to let early adopters know that they should use `v1`

joshmgross:
>The plan all along was to put dist/index.js into master as well as the releases branch so there should be no deprecation or breaking changes. 

I see, I missed it in the big set of files in the PR 👍

konradpabjan:
> Have you run a test using `setup-python@konradpabjan/ncc`?

Yup, see the screenshot attached in the top description 😃
Full run located here: https://github.com/bbq-beets/konradpabjan-testing/runs/290026972#step:3:1

PR #46, opened 2019-12-27
Update terminology in setup-python

brcrista:
Some of the comments are using outdated terminology.  I also took the chance to touch up the contributor guidelines.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -4,7 +4,7 @@ description: 'Set up a specific version of Python and add the command-line tools
 author: 'GitHub'
 inputs:
   python-version:
-    description: 'Version range or exact version of a Python version to use, using SemVer's version range syntax.'
+    description: "Version range or exact version of a Python version to use, using SemVer's version range syntax."
     default: '3.x'
   architecture:
     description: 'The target architecture (x86, x64) of the Python interpreter.'
--- a/action.yml
+++ b/action.yml
@@ -1,9 +1,10 @@
+---
 name: 'Setup Python'
-description: 'Set up a specific version of Python and add the command-line tools to the PATH'
+description: 'Set up a specific version of Python and add the command-line tools to the PATH.'
 author: 'GitHub'
 inputs:
   python-version:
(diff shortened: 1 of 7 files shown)


dist/index.js
```diff
@@ -34,7 +34,7 @@ module.exports =
 /******/ 	// the startup function
 /******/ 	function startup() {
 /******/ 		// Load entry module and return exports
-/******/ 		return __webpack_require__(429);
+/******/ 		return __webpack_require__(264);
```
brcrista: I ran with version `0.20.5` of `ncc`.

action.yml
```diff
@@ -1,9 +1,10 @@
+---
 name: 'Setup Python'
-description: 'Set up a specific version of Python and add the command-line tools to the PATH'
+description: 'Set up a specific version of Python and add the command-line tools to the PATH.'
 author: 'GitHub'
 inputs:
   python-version:
-    description: 'Version range or exact version of a Python version to use, using semvers version range syntax.'
+    description: 'Version range or exact version of a Python version to use, using SemVer's version range syntax.'
```
brcrista: See https://semver.org/#semantic-versioning-specification-semver

yaml-lint-config.yml
```diff
@@ -1,7 +1,7 @@
 extends: default
 
 rules:
-  # 80 chars should be enough, but don't fail if a line is longer
+  # 120 chars should be enough, but don't fail if a line is longer
```
brcrista: I had tried spreading the description, etc., across multiple lines, but I decided it would probably be better to raise the limit.  Monitors have gotten bigger since the days of the 80-char guideline, so I think 120 is a good number today.

brcrista:
@konradpabjan I just noticed we don't run the [main workflow](https://github.com/actions/setup-python/blob/master/.github/workflows/workflow.yml) on pull requests.  Do you want me to enable it?

brcrista:
@konradpabjan can you merge?  I don't have write access to this repo.

Issue #50, opened 2020-01-07
Show installed version in action log


robindaumann:
I use `actions/setup-python@v1` in my project with a version constraint like this `python-version: '3.x'`. In the actions log i can not see the actually installed version, like it is the case for `actions/setup-elixir@v1.2.0` where I also use a version constraint. It would be nice if we could extend the output.

Output for this action:
```
Run actions/setup-python@v1
  with:
    python-version: 3.x
    architecture: x64
Added matchers: 'python'. Problem matchers scan action output for known warning or error strings and report these inline.
```

Output for setup-elixir:
```
Run actions/setup-elixir@v1.2.0
  with:
    otp-version: 22.x
    elixir-version: 1.9.x
    install-hex: true
    install-rebar: true
Installing OTP 22.2.1
Installing Elixir v1.9.4 <--- this line
```

konradpabjan:
PR #51 has been merged in with the corresponding changes.

All changes from `master` have also been merged into `releases/v1` and tags have been updated. Using `setup-python@v1` will now show the version that was configured

Thanks again for the contribution @robindaumann

PR #51, opened 2020-01-08
output installed version number after setup

robindaumann:
implements #50 

Example run with changes: [Actions](https://github.com/robindaumann/potential-barnacle/commit/cdc009b48e76c64438854612a36e8c07a46ac041/checks?check_suite_id=390026563)

### DIFF
--- a/src/setup-python.ts
+++ b/src/setup-python.ts
@@ -7,7 +7,8 @@ async function run() {
     let version = core.getInput('python-version');
     if (version) {
       const arch: string = core.getInput('architecture', {required: true});
-      await finder.findPythonVersion(version, arch);
+      let installed = await finder.findPythonVersion(version, arch);
+      console.log(`Successfully setup ${installed}.`);
     }
     const matchersPath = path.join(__dirname, '..', '.github');
     console.log(`##[add-matcher]${path.join(matchersPath, 'python.json')}`);
--- a/src/setup-python.ts
+++ b/src/setup-python.ts
@@ -7,8 +7,10 @@ async function run() {
     let version = core.getInput('python-version');
     if (version) {
       const arch: string = core.getInput('architecture', {required: true});
-      let installed = await finder.findPythonVersion(version, arch);
-      console.log(`Successfully setup ${installed}.`);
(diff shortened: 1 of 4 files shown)


sburns:
I'd like to see this go a tad further and write the version as an [action output](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/metadata-syntax-for-github-actions#outputs). This way, we _should_ be able to reference that output in later steps. Specifically, I'd like to be able to build a virtualenv cache key based on the installed version of python.

robindaumann:
What format should the output have? Like in the install dir e.g. `Python/3.8.0` or like in the console output `Python 3.8.0`?

sburns:
Personally I'd prefer no slashes or spaces. `3.7.6`, `3.6.9`, `3.8.0`, `pypy2`, etc. The goal is to relay the specific version installed for cases that we ask for a non-specific version, like `3.7.X`.

robindaumann:
That was exactly my use case. I will implement the output in this pr.

src/setup-python.ts
```diff
@@ -7,7 +7,10 @@ async function run() {
     let version = core.getInput('python-version');
     if (version) {
       const arch: string = core.getInput('architecture', {required: true});
-      await finder.findPythonVersion(version, arch);
+      const installed = await finder.findPythonVersion(version, arch);
+      console.log(
+        `Successfully setup ${installed.impl} (${installed.version}).`
```
konradpabjan: The extra `.` seems unnecessary 

```suggestion
        `Successfully setup ${installed.impl} (${installed.version})`
```

konradpabjan:
Tested out your changes, looks good 👍

Better output/logging is always good and this is a pretty small addition. Example output can be seen here: https://github.com/konradpabjan/Testing2/runs/489889083?check_suite_focus=true

@brcrista @madhurig any thoughts or objections?

konradpabjan:
Thanks for the contribution @robindaumann 

I'll go ahead and merge your changes in. The small update that I suggested I'll address myself before pushing out a new release

robindaumann:
Alright thanks :)

PR #75, opened 2020-03-25
Update NPM packages

konradpabjan:
Sanity check to make sure everything works before going to master: https://github.com/konradpabjan/Testing2/runs/533477128?check_suite_focus=true

This should make the security warning go away that we're seeing. 

I used `ncu` to update all the packages (haven't found a better way to update all npm packages)
https://stackoverflow.com/questions/16073603/how-do-i-update-each-dependency-in-package-json-to-the-latest-version

Also `npm run format-check` is green on my windows box, but it's failing here?!? looks like it's just line endings so that can be ignored, still curious as to why (I've got CRLF in vscode and `core.autocrlf` set to true for git)?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -9,12 +9,12 @@ jobs:
         operating-system: [ubuntu-latest, windows-latest]
     steps:
     - name: Checkout
-      uses: actions/checkout@master
+      uses: actions/checkout@v2
 
-    - name: Set Node.js 10.x
-      uses: actions/setup-node@master
+    - name: Set Node.js 12.x
+      uses: actions/setup-node@v1
       with:
-        node-version: 10.x
+        node-version: 12.x
 
     - name: npm install
       run: npm install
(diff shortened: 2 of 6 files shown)


brcrista:
Why do we need to update all of the packages?

konradpabjan:
> Why do we need to update all of the packages?

Just going on an update spree. Dependabot has been submitting a few PRs here and there so I figure it's best to keep packages up to date: https://github.com/actions/setup-python/pull/73

brcrista:
> looks like it's just line endings so that can be ignored, still curious as to why (I've got CRLF in vscode and core.autocrlf set to true for git)?

A few tricks here to help:
- You can see the actual line endings in Notepad++ if you show all characters.
- With `core.autocrlf`, you can blow away your working tree (don't delete the `.git` directory of course) and then `git checkout -f konradpabjan/npm-updates` to force line ending normalization on all files.
- To avoid problems in the future, you can add a `.gitattributes` file at the top level of the repo with the following contents:

```
* text=auto
```

`.gitattributes` will override anyone's local Git config, so we don't have to rely on everyone having the same `core.autocrlf` settings.

konradpabjan:
@brcrista figured it out, `* text=auto` is needed along with `eol=lf`. I suspect git might have some different settings during checkout on our runners which is why this weirdness was happening.

I followed the recommendations here: https://prettier.io/docs/en/options.html#end-of-line

Issue #29, opened 2019-10-06
Support custom Python versions


SerVB:
My project requires Python 2.7.7 because it has to be compatible with a game that uses this version. It will be cool if the `setup-python` supports old versions. Now it doesn't:

```
##[error]Version 2.7.7 with arch x64 not found
Available versions:

2.7.16 (x64)
3.5.7 (x64)
3.6.9 (x64)
3.7.4 (x64)
##[error]Node run failed with exit code 1
```

So I have to add script to install it manually:

<https://github.com/SerVB/TheRecentStatOfYou/blob/cc20ecf4e2f530feb219a9e28dfd7aea847ea6a6/.github/workflows/build.yml#L9-L21>

Can you implement executing these lines if the version not in [2.7.16, 3.5.7, 3.6.9, 3.7.4]? My script will be much smaller!

konradpabjan:
We deliberately choose to only include a single major/minor version of Python in the tools cache. `2.7.16` is the most recent version of Python 2.7. Every time there is an update, we remove the old version (example going from Python 3.5.8 -> 3.5.9). 

Manually installing a specific older version of Python is the best solution we can recommend.

SerVB:
That's pity... Could you share reasons?

pcraciunoiu:
Heroku requires a python version pinned AFAIK, and so every time the latest version changes in GitHub actions, builds break for our projects. This is costing us a lot of time, and is not something I had to deal with for CircleCI. If we were to manually install, I assume it would be slower and cost more in build time? Seems to me like you can at least support a version for a month or an overlap period.

PR #82, opened 2020-04-14
Support installation specified of Python version for Linux

SerVB:
Follow-up of #29.

I've forked and added support for installing custom version if it's not found.

This implementation is MVP, so:
* only Linux is supported.
* path of installation is needed to be discussed.
* there are no tests (but I've used my forked action in my repo and it works: <https://github.com/SerVB/TheRecentStatOfYou/actions/runs/78111473>).

Can it be accepted to the main repo? I'm ready for discussions!

P.S.: Any ideas how can I fix the problem with unchanged lines marked as changed in this PR? Maybe something related to line endings...

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1291,27 +1291,32 @@ function getVariable(variableName) {
 }
 function downloadLinuxCpython(version) {
     return __awaiter(this, void 0, void 0, function* () {
-        const home = yield getVariable('HOME');
-        yield exec.exec('bash', [
-            '-c',
-            `
-    set -e # Any command which returns non-zero exit code will cause this shell script to exit immediately
-    set -x # Activate debugging to show execution details: all commands will be printed before execution
-    
-    sudo apt-get install build-essential checkinstall
-    sudo apt-get install libreadline-gplv2-dev libncursesw5-dev libssl-dev libsqlite3-dev tk-dev libgdbm-dev libc6-dev libbz2-dev
-
-    cd $HOME
-    wget https://www.python.org/ftp/python/${version}/Python-${version}.tgz
-
-    tar -xvf Python-${version}.tgz
-    cd Python-${version}
-    ./configure
-    make
-    sudo checkinstall -y
-    `
-        ]);
-        return `${home}/Python-${version}`;
+        try {
+            const home = yield getVariable('HOME');
+            yield exec.exec('bash', [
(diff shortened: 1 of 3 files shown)


SerVB:
Oops, will investigate the failed test.

SerVB:
It seems that error in `exec.exec` crashes tests. I've added `try catch` around it, but that doesn't help. Any suggestions?

SerVB:
@konradpabjan, could you help? Or advise a place to ask?

konradpabjan:
@SerVB thanks for the PR! We have something new on the way for `setup-python` that downloads specific versions of python if they are not available. This will allow non-cached versions of python to be installed and it will work with Windows Linux & Mac. It should address the problem you're facing 😀

Check out this PR: https://github.com/actions/setup-python/pull/85

Can you open an issue in [actions/python-versions](https://github.com/actions/python-versions) for the `2.7.7` version of python. 

In your example run it takes about 6 minutes to compile the specific version of Python, while downloading an already compiled package and installing it should take about 10 seconds. In `python-versions` we also compile with `--enable-optimizations` so it takes a lot longer to compile, but the binaries are little bit better so I think the best solution is to get `2.7.7` available in `python-versions`.

konradpabjan:
Sorry that we couldn't get `2.7.7` into `python-versions` 😕 The version is just too old and there were a lot of weird issues during compile time that we can't prioritize at the moment.

Based off of the consensus in https://github.com/actions/virtual-environments/issues/810 I'll close out this PR. Thanks again for initially creating this PR and apologies that we can't support your specific use case.

SerVB:
@konradpabjan , and why don't you want to add a feature to compile any python version manually, as I propose here?

I don't understand why you see only two variants:
* Precompile 2.7.7.
* Don't support 2.7.7 at all.

This PR is not even about 2.7.7, it's more general :) #85 doesn't cover it.

konradpabjan:
@SerVB 

We're trying to keep any compile logic outside of this action. Historically, where/how we compiled Python from source was closed source, however we recently moved everything to `actions/python-versions` and we pull everything from there now and any updates will take place there.

Maintaining compilation in this action and in another place is a maintenance burden that is hard to justify. It would also cause some confusion for users (do they download an already compiled version or do they have to download from source and build it themselves, or do they use an already installed version?).

Issue #71, opened 2020-03-11
Also look at `$RUNNER_TOOL_CACHE` for finding the tool cache directory


brcrista:
See this code: https://github.com/actions/runner/blob/6bec1e3bb832aad26f4ad5b64759a8e4d468df24/src/Runner.Common/HostContext.cs#L233

The runner prefers `RUNNER_TOOL_CACHE` and then falls back to `RUNNER_TOOLSDIRECTORY`.  `setup-python` only uses the latter: https://github.com/actions/setup-python/blob/9ac730844c47e52575257e93faf193c4530bc095/src/find-python.ts#L6

- [ ] Update the code
- [ ] Update the [self-hosted guidance](https://github.com/actions/setup-python#using-setup-python-with-a-self-hosted-runner)

konradpabjan:
Closing, the `RUNNER_TOOLSDIRECTORY` was recently removed as it is no longer needed. See https://github.com/actions/setup-python/pull/85#discussion_r415658230 for more info

PR #85, opened 2020-04-24
Use GitHub releases to download python versions

MaksimZhukov:
This pull-request improves `setup-python` action to add ability to download specific version of Python on flight if it is not available by default.

**Details:**
`setup-python` action will download and install specific Python version from GitHub releases ([actions/python-versions](https://github.com/actions/python-versions/releases)) in case the version is not found in the local cache. All versions of Python available for installation are published in [actions/python-versions](https://github.com/actions/python-versions) repository.
All available versions are listed in the [version-manifest.json](https://github.com/actions/python-versions/blob/master/versions-manifest.json) file.

**Installation time:**

- Ubuntu / macOS: 10-20 seconds
- Windows: ~ 1 minute (mostly related to fact that we use MSI installer for Python on Windows)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -8,11 +8,8 @@ inputs:
     default: '3.x'
   architecture:
     description: 'The target architecture (x86, x64) of the Python interpreter.'
-    default: 'x64'
   token:
-    description: >
-      Personal access token (PAT) used to fetch the manifest file from master of repository
-      [Learn more about creating and using encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
+    description: Used to pull python distributions from actions/python-versions. Since there's a default, this is typically not supplied by the user.
     default: ${{ github.token }}
 outputs:
   python-version:
--- a/action.yml
+++ b/action.yml
@@ -9,6 +9,11 @@ inputs:
   architecture:
     description: 'The target architecture (x86, x64) of the Python interpreter.'
     default: 'x64'
+  token:
+    description: >
+      Personal access token (PAT) used to fetch the manifest file from master of repository
+      [Learn more about creating and using encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
+    default: ${{ github.token }}
 outputs:
   python-version:
     description: "The installed python version. Useful when given a version range as input."
--- a/src/find-python.ts
+++ b/src/find-python.ts
@@ -83,7 +83,7 @@ async function useCpythonVersion(
     architecture
   );
   if (!installDir) {
-    core.info(`Version ${semanticVersionSpec} is not found in local cache`);
+    core.info(`Version ${semanticVersionSpec} was not found in the local cache`);
     const foundRelease = await installer.findReleaseFromManifest(
       semanticVersionSpec,
       architecture
--- a/src/find-python.ts
+++ b/src/find-python.ts
@@ -83,7 +83,9 @@ async function useCpythonVersion(
     architecture
   );
   if (!installDir) {
-    core.info(`Version ${semanticVersionSpec} was not found in the local cache`);
+    core.info(
+      `Version ${semanticVersionSpec} was not found in the local cache`
+    );
     const foundRelease = await installer.findReleaseFromManifest(
       semanticVersionSpec,
       architecture
--- a/src/find-python.ts
+++ b/src/find-python.ts
@@ -3,6 +3,8 @@ import * as path from 'path';
 
 import * as semver from 'semver';
 
+import * as installer from './install-python';
+
 let cacheDirectory = process.env['RUNNER_TOOLSDIRECTORY'] || '';
 
 if (!cacheDirectory) {
@@ -92,29 +94,31 @@ async function useCpythonVersion(
   const semanticVersionSpec = pythonVersionToSemantic(desugaredVersionSpec);
   core.debug(`Semantic version spec of ${version} is ${semanticVersionSpec}`);
 
-  const installDir: string | null = tc.find(
+  let installDir: string | null = tc.find(
     'Python',
     semanticVersionSpec,
     architecture
   );
   if (!installDir) {
-    // Fail and list available versions
-    const x86Versions = tc
-      .findAllVersions('Python', 'x86')
-      .map(s => `${s} (x86)`)
-      .join(os.EOL);
+    core.info(`Version ${semanticVersionSpec} is not found in local cache`);
+    const foundRelease = await installer.findReleaseFromManifest(
+      semanticVersionSpec,
+      architecture
+    );
 
-    const x64Versions = tc
-      .findAllVersions('Python', 'x64')
-      .map(s => `${s} (x64)`)
-      .join(os.EOL);
+    if (foundRelease && foundRelease.files && foundRelease.files.length > 0) {
+      core.info(`Version ${semanticVersionSpec} is available for downloading`);
+      await installer.installCpythonFromRelease(foundRelease);
 
+      installDir = tc.find('Python', semanticVersionSpec, architecture);
+    }
+  }
+
+  if (!installDir) {
     throw new Error(
       [
         `Version ${version} with arch ${architecture} not found`,
-        'Available versions:',
-        x86Versions,
-        x64Versions
+        `The list of all available versions can be found here: ${installer.MANIFEST_URL}`
       ].join(os.EOL)
     );
   }
--- a/src/find-python.ts
+++ b/src/find-python.ts
@@ -5,23 +5,6 @@ import * as semver from 'semver';
 
 import * as installer from './install-python';
 
-let cacheDirectory = process.env['RUNNER_TOOLSDIRECTORY'] || '';
-
-if (!cacheDirectory) {
-  let baseLocation;
-  if (process.platform === 'win32') {
-    // On windows use the USERPROFILE env variable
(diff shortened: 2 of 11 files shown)


action.yml
```diff
@@ -9,6 +9,11 @@ inputs:
   architecture:
     description: 'The target architecture (x86, x64) of the Python interpreter.'
     default: 'x64'
+  token:
+    description: >
+      Personal access token (PAT) used to fetch the manifest file from master of repository
```
bryanmacfarlane, line 14: Note that the github.token is not a PAT and we should not encourage this to be set to a PAT.  We shouldn't even encourage it to be set.  It's just a way to declaratively say that the action needs access to the per job.  See the description I used in the setup-node PR

src/install-python.ts
```diff
@@ -0,0 +1,64 @@
+import * as path from 'path';
+import * as core from '@actions/core';
+import * as tc from '@actions/tool-cache';
+import * as exec from '@actions/exec';
+import {ExecOptions} from '@actions/exec/lib/interfaces';
+
+const AUTH_TOKEN = core.getInput('token');
+const MANIFEST_REPO_OWNER = 'actions';
+const MANIFEST_REPO_NAME = 'python-versions';
+export const MANIFEST_URL = `https://raw.githubusercontent.com/${MANIFEST_REPO_OWNER}/${MANIFEST_REPO_NAME}/master/versions-manifest.json`;
```
bryanmacfarlane, line 10: This should be no longer needed.

src/install-python.ts
```diff
@@ -0,0 +1,64 @@
...
+        core.debug(data.toString().trim());
+      }
+    }
+  };
+
+  core.info('Execute installation script');
+  if (IS_WINDOWS) {
+    await exec.exec('powershell', ['./setup.ps1'], options);
+  } else if (IS_LINUX) {
+    await exec.exec('sudo', ['bash', './setup.sh'], options);
```
bryanmacfarlane, line 60: I think this will work fine on hosted machines but self hosted may prompt?

src/install-python.ts
```diff
@@ -0,0 +1,64 @@
...
+  core.info('Extract downloaded archive');
+  let pythonExtractedFolder;
+  if (IS_WINDOWS) {
+    pythonExtractedFolder = await tc.extractZip(pythonPath, `./${fileName}`);
+  } else {
+    pythonExtractedFolder = await tc.extractTar(pythonPath, `./${fileName}`);
+  }
+
+  const options: ExecOptions = {
+    cwd: pythonExtractedFolder,
```
bryanmacfarlane, line 47: Side question on the extracted folder.  After download and extraction is done, is the physical layout in the toolcache the same?  Or at least close enough where it won't be problematic between v1 and v2?  I'm thinking of a org wide self hosted runner (just shipped) and one workflow runs python-setup@v2 and another runs python-setup@v1 and get's a folder hit.  Another variant of that same scenario is if they try v2 or v2-beta and then go back to v1.  The mechanism of how it get's there doesn't matter (downloaded, script to extract etc) but the basic binary layout of what is added to the PATH needs to be the same.

src/install-python.ts
```diff
@@ -0,0 +1,64 @@
...
+    manifest,
+    architecture
+  );
+}
+
+export async function installCpythonFromRelease(release: tc.IToolRelease) {
+  const downloadUrl = release.files[0].download_url;
+
+  core.info(`Download from "${downloadUrl}"`);
+  const pythonPath = await tc.downloadTool(downloadUrl, undefined, AUTH_TOKEN);
```
bryanmacfarlane, line 36: nice.  thanks for passing the token

src/find-python.ts
```diff
@@ -3,6 +3,8 @@ import * as path from 'path';
 
 import * as semver from 'semver';
 
+import * as installer from './install-python';
+
 let cacheDirectory = process.env['RUNNER_TOOLSDIRECTORY'] || '';
```
bryanmacfarlane, line 8: This is an old variable and why do you need it?  It seems like you're breaking encapsulation.  The runner will always set it and that's an internal implementation detail of tool-cache.

action.yml
```diff
@@ -9,6 +9,11 @@ inputs:
   architecture:
     description: 'The target architecture (x86, x64) of the Python interpreter.'
     default: 'x64'
```
bryanmacfarlane, line 11: I think the code for architecture should be 
```javascript
let arch = core.getInput("architecture") | os.arch()
```
and the action yml should have no default.

That way, if the user doesn't supply the optional input, the most likely outcome will "just work".  It's only when they want to do something explicitly different like run i386 python on x64 hardware where they need to be explicit. 

Without that, you might see a bunch of copy / paste issues where it leads to hard to diagnose issues.

On a related note, does the action do validation for combinations that will not work?

src/install-python.ts
brcrista: I very vaguely remember some issue from AzP where `exec('sudo', ...)` caused problems.  @bryanmacfarlane does that ring a bell?

brcrista: Oh, never mind -- it was running `conda` as `sudo` would cause problems.  So, specific to a particular toolchain.

bryanmacfarlane: Yeah, hosted runs as config-less sudo so this will just work but I suspect on self hosted, it may not and you'll get prompted.  I'm not sure if there's a way to check and error saying you need that.

src/install-python.ts
maxim-lobanov: We need the url for error message:
```
Version ${version} with arch ${architecture} not found
Available versions: The list of all available versions can be found here: ${installer.MANIFEST_URL}
```

src/install-python.ts
maxim-lobanov: After download, extraction and invocation of installation script, tool layout is pretty same like V1 had. No difference at all

src/install-python.ts
maxim-lobanov: I think we can use `await exec.exec('sudo', ['bash', './setup.sh'], options);`
`-n, --non-interactive         non-interactive mode, no prompts are used`

So if sudo requires password, command will fail immediately with `sudo: a password is required`
cc: @MaksimZhukov

konradpabjan:
The `README` could use an update with regards to some of the new changes introduced here. A few things off the top of my head:

- An example with the new `token` input parameter
- An update to the `available-versions-of-python` section and `hosted-tool-cache`  section with some clarifications around what is cached and what can be downloaded
- Some mention of our new fancy `actions/python-versions` repo 😀

I also think some it would be good to have some information about what happens when a new version of Python is released (for example if `3.8.3` is released and replaces version `3.8.2`). What can users expect to happen if they have `3.8` specified and what should they do if they need a specific version.

src/find-python.ts
konradpabjan: We have an issue for this variable: https://github.com/actions/setup-python/issues/71

maxim-lobanov: @konradpabjan , we don't find any occurrence where this variable `cacheDirectory` is used.
Please see [this link for details](https://github.com/actions/setup-python/search?q=cacheDirectory&unscoped_q=cacheDirectory).
In our understanding, it can be safety removed.

konradpabjan: Well it looks like it can be safety removed! Earlier when the issue was filled I guess @brcrista and I didn't check that this was actually being used.

Lets get rid of it 😀

action.yml
MaksimZhukov: Thanks for your point!

src/install-python.ts
MaksimZhukov: Thanks, I've added the `-n` flag

src/find-python.ts
MaksimZhukov: Thanks, I've removed the `cacheDirectory` variable

action.yml
maxim-lobanov: Updated PR to use `let arch = core.getInput("architecture") || os.arch()`
In this case, if customer uses architecture that is not supported by our binaries, he will get error and will have to override default architecture

bryanmacfarlane:
Note that concerning new releases, until the cache is update it will match against that first and we will have to be good about staying on top of it - ideally automated.  In node tool in azure pipelines, we had an input which was always check source first (read through to dist).  Can't do that with python since we build.  Also note that an option like that says I must have the very latest at the cost of the instability

src/install-python.ts
MaksimZhukov: We've fixed permissions issue on Ubuntu. It looks like `sudo` is no longer required, so I've removed linux condition

src/find-python.ts
```diff
@@ -92,29 +77,31 @@ async function useCpythonVersion(
...
     semanticVersionSpec,
     architecture
   );
   if (!installDir) {
-    // Fail and list available versions
-    const x86Versions = tc
-      .findAllVersions('Python', 'x86')
-      .map(s => `${s} (x86)`)
-      .join(os.EOL);
+    core.info(`Version ${semanticVersionSpec} is not found in local cache`);
```
konradpabjan, line 86:
```suggestion
    core.info(`Version ${semanticVersionSpec} was not found in the local cache`);
```

MaksimZhukov: Thank you!

konradpabjan:
I'll create a new tag and release tomorrow (`v1.3.0`). There are a significant number of users pinned to `@master` so if there are any issues overnight we should hear about it. I don't expect anything to break or any compat issues since we've done a lot of testing.

PR #87, opened 2020-04-29
README Updates

konradpabjan:
Updates to the `README` that should be made as part of https://github.com/actions/setup-python/pull/85

Tracking issue: https://github.com/github/c2c-actions-service/issues/1025

Rendered 🖌 [README](https://github.com/actions/setup-python/blob/f9a2b4d711c0beaea6accfc56914e5dca4290bc4/README.md) 🎨

I didn't include any mention of the new `token` input since it's effectively an implementation detail. Users shouldn't need to do anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Check out our detailed guide on using [Python with GitHub Actions](https://help.
     - For detailed information regarding the available versions of Python that are installed see [Software installed on GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners).
     - For every minor version of Python, expect only the latest patch to be preinstalled
     - If `3.8.1` is installed for example, and `3.8.2` is released, expect `3.8.1` to be removed and replaced by `3.8.2` in the tools cache
-    - If the exact patch version doesn't matter to you, specifying just the major and minor version will get you the latest preinstalled patch version. In the previous example, the version spec 3.8 will use the 3.8.2 python version found in the cache
+    - If the exact patch version doesn't matter to you, specifying just the major and minor version will get you the latest preinstalled patch version. In the previous example, the version spec `3.8` will use the `3.8.2` Python version found in the cache
 - Downloadable Python versions from GitHub Releases ([actions/python-versions](https://github.com/actions/python-versions/releases))
     - All available versions are listed in the [version-manifest.json](https://github.com/actions/python-versions/blob/master/versions-manifest.json) file
     - If there is a specific version of Python that is not available, you can open an issue in the `python-versions` repository 
@@ -120,8 +120,8 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
 # Specifying a Python version
 
 If there is a specific version of Python that you need and you don't want to worry about any potential breaking changes due to patch updates (going from `3.7.5` to `3.7.6` for example), you should specify the exact major, minor, and patch version (such as `3.7.5`)
-  - The only downside to this is that setup will take a little longer since the exact version will have to be downloaded if the exact version is not already installed on the runner due to more recent versions
-  - MSI installers are used on Windows for this, so runs will take a little longer to setup vs Mac and Linux
+  - The only downside to this is that set up will take a little longer since the exact version will have to be downloaded if the exact version is not already installed on the runner due to more recent versions
+  - MSI installers are used on Windows for this, so runs will take a little longer to set up vs Mac and Linux
 
 You should specify only a major and minor version if you are okay with the most recent patch version being used
   - For example, if you just want a `3.8` version of Python, and it doesn't matter if `3.8.1` is used or the more recent `3.8.2`,
@@ -131,13 +131,13 @@ You should specify only a major and minor version if you are okay with the most 
 
 # Using `setup-python` with a self hosted runner
 
-If you would like to use `setup-python` and a self-hosted runner, there isn't much that you need to do. When `setup-python` is run for the first time with a version of Python that it doesn't have, it will download the appropriate version, and setup the tools cache on your machine. Any subsequent runs will use the python versions that were previously downloaded.
+If you would like to use `setup-python` and a self-hosted runner, there isn't much that you need to do. When `setup-python` is run for the first time with a version of Python that it doesn't have, it will download the appropriate version, and set up the tools cache on your machine. Any subsequent runs will use the Python versions that were previously downloaded.
 
 A few things to look out for when `setup-python` is first setting up the tools cache
 - If using Windows, your runner needs to be running as an administrator so that the appropriate directories and files can be setup. On Linux and Mac, you also need to be running with elevated permissions
 - On Windows, you need `7zip` installed and added to your `PATH` so that files can be extracted properly during setup
 - MSI installers are used when setting up Python on Windows. A word of caution as MSI installers update registry settings
-- The 3.8 MSI installer for Windows will not let you install another 3.8 version of Python. If `setup-python` fails for a 3.8 version of Python, make sure any previously installed versions are removed by going to "Apps and Features" in the Windows settings and uninstalling any
+- The 3.8 MSI installer for Windows will not let you install another 3.8 version of Python. If `setup-python` fails for a 3.8 version of Python, make sure any previously installed versions are removed by going to "Apps & Features" in the Settings app.
 
 # Using Python without `setup-python`
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ jobs:
         run: python -c "import sys; print(sys.version)"
 ```
 
-Download and setup a version of Python that does not come pre-installed on an image:
+Download and set up a version of Python that does not come preinstalled on an image:
 ```yaml
 jobs:
   build:
@@ -97,6 +97,7 @@ Check out our detailed guide on using [Python with GitHub Actions](https://help.
     - For detailed information regarding the available versions of Python that are installed see [Software installed on GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners)
     - For every minor version of Python, expect only the latest patch to be preinstalled. See [Semantic Versioning](https://semver.org/) for more information
     - If `3.8.1` is installed for example, and `3.8.2` is released, expect `3.8.1` to be removed and replaced by `3.8.2` in the tools cache
+    - If the exact patch version doesn't matter to you, specifying just the major and minor version will get you the latest preinstalled patch version. In this case, for example, the version spec 3.8 will use the 3.8.2 found in the cache
 - Downloadable Python versions from GitHub Releases ([actions/python-versions](https://github.com/actions/python-versions/releases))
     - All available versions are listed in the [version-manifest.json](https://github.com/actions/python-versions/blob/master/versions-manifest.json) file
     - If there is a specific version of Python that is not available, you can open an issue in the `python-versions` repository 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,13 +69,41 @@ jobs:
         run: python -c "import sys; print(sys.version)"
 ```
 
+Download and setup a version of Python that does not come pre-installed on an image:
+```yaml
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+        python-version: [3.5.4, 3.6.6, 3.7.4, 3.8.1]
```
brcrista, line 78: This change should address the problem of sliding patch versions, but I still think people shouldn't pin to a patch version unless they have a good reason.  Patch versions are bugfix-only.

What do you think of:

```suggestion
        python-version: [3.5, 3.6, 3.7.4, 3.8]
```

README.md
```diff
@@ -67,13 +69,41 @@ jobs:
         run: python -c "import sys; print(sys.version)"
 ```
 
+Download and setup a version of Python that does not come pre-installed on an image:
```
brcrista, line 72: `set up`

A good trick for this: can you move `up` around in the sentence and have it still make sense?

```
Download and set a version of Python up
```

vs.

```
My .bashrc defines my shell setup
```

brcrista, line 72: Above, you used `preinstalled` (no hyphen).  I think either one is fine.

README.md
```diff
@@ -67,13 +69,41 @@ jobs:
...
 
 Check out our detailed guide on using [Python with GitHub Actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-python-with-github-actions).
 
+# Available versions of Python
+
+`setup-python` is able to configure Python from two sources
+
+- Preinstalled versions of Python in the tools cache on GitHub-hosted runners
+    - For detailed information regarding the available versions of Python that are installed see [Software installed on GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners)
+    - For every minor version of Python, expect only the latest patch to be preinstalled. See [Semantic Versioning](https://semver.org/) for more information
```
brcrista, line 98: +1

README.md
```diff
@@ -67,13 +69,41 @@ jobs:
...
+    - If `3.8.1` is installed for example, and `3.8.2` is released, expect `3.8.1` to be removed and replaced by `3.8.2` in the tools cache
```
brcrista, line 99: The upshot of this is basically that your build will take a little longer, since it has to download the Python versions.

You might want to add, "If the exact patch version doesn't matter to you, specifying just the major and minor version will get you the latest preinstalled patch version.  In this case, for example, the version spec `3.8` will use the `3.8.2` found in the cache.

README.md
```diff
@@ -87,9 +117,18 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
 
 # Using `setup-python` with a self hosted runner
 
-If you would like to use `setup-python` on a self-hosted runner, you will need to download all versions of Python & PyPy that you would like and setup a similar tools cache locally for your runner.
+If you would like to use `setup-python` and a self-hosted runner, you have two options
+  - Setup a tools cache locally and download all the versions of Python & PyPy that you would like once
```
brcrista, line 121:
```suggestion
  - Set up a tools cache locally and install all the versions of Python and PyPy that you need there
```

brcrista, line 121:
```suggestion
  - Set up a tools cache locally and install all the versions of Python and PyPy that you need there
```

README.md
```diff
@@ -87,9 +117,18 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
 
 # Using `setup-python` with a self hosted runner
 
-If you would like to use `setup-python` on a self-hosted runner, you will need to download all versions of Python & PyPy that you would like and setup a similar tools cache locally for your runner.
+If you would like to use `setup-python` and a self-hosted runner, you have two options
```
brcrista, line 120: This is good content; I like showing the pros and cons.  I'll Slack you a proposed rewording of this section to make it more clear.

brcrista: Actually, once they run `setup-python` once on their self-hosted runner, the versions they download will still be in the tools cache on subsequent runs.  So, let's skip the part about setting it up themselves.

```md
When using `setup-python` on a self-hosted runner, it will need to download the versions of Python it needs the first time it runs on that runner.  For subsequent runs, it will use the cached versions.
```

A technical point: say a self-hosted runner downloads `3.8.2`, and then `3.8.3` is released.  I suppose the self-hosted runner will be stuck on `3.8.2` until they delete it from the cache.

README.md
```diff
@@ -106,14 +145,21 @@ If you would like to use `setup-python` on a self-hosted runner, you will need t
 
 `setup-python` helps keep your dependencies explicit and ensures consistent behavior between different runners. If you use `python` in a shell on a GitHub hosted runner without `setup-python` it will default to whatever is in PATH. The default version of Python in PATH vary between runners and can change unexpectedly so we recommend you always use `setup-python`.
 
-# Available versions of Python
+# Need to open an issue?
+
+Python versions that we compile and setup can be found in the [actions/python-versions](https://github.com/actions/python-versions) repository. You can see the build scripts, configurations, and everything that is used. You should open an issue in the `python-versions` repository if:
```
brcrista, line 150: Well, they haven't been set up until `setup-python` runs 😉 

```suggestion
Python versions available for `setup-python` can be found in the [actions/python-versions](https://github.com/actions/python-versions) repository. You can see the build scripts, configurations, and everything that is used. You should open an issue in the `python-versions` repository if:
```

README.md
```diff
@@ -106,14 +145,21 @@ If you would like to use `setup-python` on a self-hosted runner, you will need t
...
+  - something might be compiled incorrectly
+  - certain modules might be missing
+  - there is a version of Python that you would like that is currently not available
+
+If you suspect something might be wrong with the tools cache or how Python gets installed on GitHub hosted runners, please open an issue in [actions/virtual-environments](https://github.com/actions/virtual-environments)
```
brcrista, line 155:
```suggestion
If you suspect something might be wrong with the tools cache or how Python gets installed on GitHub hosted runners, please open an issue in [actions/virtual-environments](https://github.com/actions/virtual-environments).
```

README.md
```diff
@@ -106,14 +145,21 @@ If you would like to use `setup-python` on a self-hosted runner, you will need t
...
 
-For detailed information regarding the available versions of Python that are installed see [Software installed on GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners)
+Any remaining issues can be filed in this repository
```
brcrista, line 157:
```suggestion
Any remaining issues can be filed in this repository.
```

README.md
```diff
@@ -67,13 +69,41 @@ jobs:
...
+
+```
+
 # Getting started with Python + Actions
 
 Check out our detailed guide on using [Python with GitHub Actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-python-with-github-actions).
 
+# Available versions of Python
+
+`setup-python` is able to configure Python from two sources
```
brcrista, line 94:
```suggestion
`setup-python` is able to configure Python from two sources:
```

README.md
```diff
@@ -67,13 +69,41 @@ jobs:
...
+
+- Preinstalled versions of Python in the tools cache on GitHub-hosted runners
+    - For detailed information regarding the available versions of Python that are installed see [Software installed on GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners)
```
brcrista, line 97:
```suggestion
    - For detailed information regarding the available versions of Python that are installed see [Software installed on GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners).
```

README.md
brcrista, line 98:
```suggestion
    - For every minor version of Python, expect only the latest patch to be preinstalled.
```

Python technically doesn't use Semantic Versioning (though it is very close).  Regardless, I think we can expect the reader to know what a patch version is.

I do think it's worth having a section explaining how the action chooses a version to use.

```markdown
The `python-version` argument uses [the semver version spec syntax](https://semver.org/) to find a version.  If a version matching the version spec is found in the tools cache, that version will be used.  If no such version is found, `setup-python` will check for an available version to download from `actions/python-versions`.  If no matching version can be found, the `setup-python` will fail.

Versions in the tools cache are updated when new versions of the virtual environments are released.
```

README.md
```diff
@@ -87,9 +117,18 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
...
+  - Setup a tools cache locally and download all the versions of Python & PyPy that you would like once
+      - Takes a little bit of time to initially setup
+      - This will be the most stable and fastest option long-term as it will require no extra downloads every-time there is a run
+  - Download and setup a version of python every-time
+      - Requires no extra setup (good if you want to quickly get up and running, discouraged for long term use)
+      - `setup-python` will take a little longer to run
+      - Note: when downloading versions of Python for Windows, an MSI installer is used which can modify some registry settings
 
-- Create an global environment variable called `AGENT_TOOLSDIRECTORY` that will point to the root directory of where you want the tools installed. The env variable is preferrably global as it must be set in the shell that will install the tools cache, along with the shell that the runner will be using.
+### Setting up a local tools cache
```
brcrista, line 129: Since the action will fill the cache itself now, I think we can delete this whole section 🎉

konradpabjan: I'll run a quick test locally (I probably should have done that first), if it works I guess we can nuke this entire section 😀 (even possibly some extra stuff above too)

README.md
konradpabjan: I'm going to test this quickly...

README.md
konradpabjan: Rip, doesn't quite work.. https://github.com/konradpabjan/Testing2/runs/629781394

README.md
```diff
@@ -106,14 +145,21 @@ If you would like to use `setup-python` on a self-hosted runner, you will need t
...
+
+Python versions available for `setup-python` can be found in the [actions/python-versions](https://github.com/actions/python-versions) repository. You can see the build scripts, configurations, and everything that is used. You should open an issue in the `python-versions` repository if:
+  - something might be compiled incorrectly
+  - certain modules might be missing
+  - there is a version of Python that you would like that is currently not available
+
+If you suspect something might be wrong with the tools cache or how Python gets installed on GitHub hosted runners, please open an issue in [actions/virtual-environments](https://github.com/actions/virtual-environments).
 
-For detailed information regarding the available versions of Python that are installed see [Software installed on GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners)
+Any remaining issues can be filed in this repository.
```
maxim-lobanov, line 157: cc: @alepauly that we have discussed today on meeting.

@konradpabjan , we have discussed the following idea:
1) Restrict creation of new issues in `python-versions` repository
2) Notify customers that they should log issues in `setup-python` repository.
3) Your team will reassign approved work items (that we want to implement) to our team if needs.

README.md
konradpabjan: Got it working... Python 3.8 has some issues on Windows, I added some information about it on the `README`

README.md
konradpabjan: I updated the instructions after testing this locally. This section should be good now and the users doesn't need to do anything special other than watch out for a few things

README.md
konradpabjan: Based on our slack discussion, I've removed the section about `virtual-environments`

README.md
konradpabjan: I added a section called `Specifying a Python version` that outlines some information about which version gets used and when there is a download

README.md
```diff
@@ -67,13 +69,43 @@ jobs:
...
 
 Check out our detailed guide on using [Python with GitHub Actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-python-with-github-actions).
 
+# Available versions of Python
+
+`setup-python` is able to configure Python from two sources:
+
+- Preinstalled versions of Python in the tools cache on GitHub-hosted runners
+    - For detailed information regarding the available versions of Python that are installed see [Software installed on GitHub-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners).
+    - For every minor version of Python, expect only the latest patch to be preinstalled
```
brcrista, line 99: Small nit, let's add periods for all bullet points that are complete sentences.  I feel like it reads better.

README.md
```diff
@@ -67,13 +69,43 @@ jobs:
...
+    - If `3.8.1` is installed for example, and `3.8.2` is released, expect `3.8.1` to be removed and replaced by `3.8.2` in the tools cache
+    - If the exact patch version doesn't matter to you, specifying just the major and minor version will get you the latest preinstalled patch version. In the previous example, the version spec 3.8 will use the 3.8.2 python version found in the cache
```
brcrista, line 101:
```suggestion
    - If the exact patch version doesn't matter to you, specifying just the major and minor version will get you the latest preinstalled patch version. In the previous example, the version spec `3.8` will use the `3.8.2` Python version found in the cache
```

README.md
```diff
@@ -85,35 +117,45 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
 - [Tools cache setup for Ubuntu](https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/hosted-tool-cache.sh)
 - [Tools cache setup for Windows](https://github.com/actions/virtual-environments/blob/master/images/win/scripts/Installers/Download-ToolCache.ps1)
 
+# Specifying a Python version
+
+If there is a specific version of Python that you need and you don't want to worry about any potential breaking changes due to patch updates (going from `3.7.5` to `3.7.6` for example), you should specify the exact major, minor, and patch version (such as `3.7.5`)
+  - The only downside to this is that setup will take a little longer since the exact version will have to be downloaded if the exact version is not already installed on the runner due to more recent versions
```
brcrista, line 123:
```suggestion
  - The only downside to this is that set up will take a little longer since the exact version will have to be downloaded if the exact version is not already installed on the runner due to more recent versions
```

README.md
```diff
@@ -85,35 +117,45 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
...
+  - MSI installers are used on Windows for this, so runs will take a little longer to setup vs Mac and Linux
```
brcrista, line 124:
```suggestion
  - MSI installers are used on Windows for this, so runs will take a little longer to set up vs Mac and Linux
```

konradpabjan: Rip, I don't learn...

README.md
```diff
@@ -85,35 +117,45 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
...
+
+You should specify only a major and minor version if you are okay with the most recent patch version being used
+  - For example, if you just want a `3.8` version of Python, and it doesn't matter if `3.8.1` is used or the more recent `3.8.2`,
```
brcrista, line 127: The bullets don't read right to me with the comma at the end.  I like bullets, but maybe this can just be a regular paragraph.

README.md
```diff
@@ -85,35 +117,45 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
...
+  - There will be a single patch version already installed on each runner for every minor version of Python that is supported
+  - The patch version that will be preinstalled, will generally be the latest and every time there is a new patch released, the older version that is preinstalled will be replaced
+  - Using the most recent patch version will result in a very quick setup since no downloads will be required since a locally installed version Python on the runner will be used
```
brcrista, line 130: This basically reiterates the above paragraph; consider removing.

README.md
```diff
@@ -85,35 +117,45 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
...
-    - Once configured, download a specific package by calling `npm install`. Note (if downloading a PyPy package on Windows, you will need 7zip installed along with `7z.exe` added to your PATH)
-- Each NPM package has multiple versions that determine the version of Python or PyPy that should be installed. 
-    - `npm install @actions/toolcache-python-windows-x64@3.7.61579791175` for example installs Python 3.7.6 while `npm install @actions/toolcache-python-windows-x64@3.6.81579791177` installs Python 3.6.8
-    - You can browse and find all available versions of a package by searching the GitHub Actions NPM registry
-![image](https://user-images.githubusercontent.com/16109154/76194005-87aeb400-61e5-11ea-9b21-ef9111247f84.png)
+If you would like to use `setup-python` and a self-hosted runner, there isn't much that you need to do. When `setup-python` is run for the first time with a version of Python that it doesn't have, it will download the appropriate version, and setup the tools cache on your machine. Any subsequent runs will use the python versions that were previously downloaded.
```
brcrista, line 134:
```suggestion
If you would like to use `setup-python` and a self-hosted runner, there isn't much that you need to do. When `setup-python` is run for the first time with a version of Python that it doesn't have, it will download the appropriate version, and set up the tools cache on your machine. Any subsequent runs will use the Python versions that were previously downloaded.
```

README.md
```diff
@@ -85,35 +117,45 @@ GitHub virtual environments are setup in [actions/virtual-environments](https://
...
+
+A few things to look out for when `setup-python` is first setting up the tools cache
+- If using Windows, your runner needs to be running as an administrator so that the appropriate directories and files can be setup. On Linux and Mac, you also need to be running with elevated permissions
+- On Windows, you need `7zip` installed and added to your `PATH` so that files can be extracted properly during setup
+- MSI installers are used when setting up Python on Windows. A word of caution as MSI installers update registry settings
+- The 3.8 MSI installer for Windows will not let you install another 3.8 version of Python. If `setup-python` fails for a 3.8 version of Python, make sure any previously installed versions are removed by going to "Apps and Features" in the Windows settings and uninstalling any
```
brcrista, line 140:
```suggestion
- The 3.8 MSI installer for Windows will not let you install another 3.8 version of Python. If `setup-python` fails for a 3.8 version of Python, make sure any previously installed versions are removed by going to "Apps & Features" in the Settings app.
```

If there's a good way for sysadmins to be able to do this through a script, that would be even better.

README.md
konradpabjan: Removed

README.md
konradpabjan: I removed the bullet point right below, so it should be better to read now

konradpabjan:
Merging this in. There are a few failing tests, but i talked to @maxim-lobanov and those all be fixed behind the scenes

PR #95, opened 2020-05-20
GHES Support

konradpabjan:
# Overview

- Bump `@actions/tool-cache` to the latest `1.5.5` release
- Largely duplicating the same work that was done for `setup-node` to support GHES in the following PR: https://github.com/actions/setup-node/pull/157


bryanmacfarlane:
LGTM.  FYI for @ericsciple

PR #112, opened 2020-07-15
Add support of pre-release Python versions

MaksimZhukov:
In scope of this PR we've added ability to set up pre-release Python versions.

**Details:**
There are several ways to specify pre-release versions, for example:
- `3.9.0-beta.4` to set up an exact pre-release version 
- `3.9.0-alpha - 3.9.0` to set up the latest available version of Python (including both pre-release and stable versions). 

Currently, we do not preinstall pre-release versions on our images. All available pre-release versions will be downloaded and installed in runtime from [versions-manifest.json](https://github.com/actions/python-versions/blob/main/versions-manifest.json) located in the [actions/python-versions](https://github.com/actions/python-versions) repository.

Also we've updated tests to cover set up of pre-release versions.


src/find-python.ts
```diff
@@ -171,14 +173,15 @@ export function pythonVersionToSemantic(versionSpec: string) {
 
 export async function findPythonVersion(
```
maxim-lobanov, line 174: @bryanmacfarlane , @konradpabjan , there is a function `pythonVersionToSemantic` that converts version from format `3.7.0b2` to `3.7.0-b2`. Not sure why this function was needed before because we have never added pre-release versions previously (may be self-hosted?). May be it makes sense to modify this function to convert from `3.7.0b2` to `3.7.0-beta.2` so version will be resolved from manifest?

https://github.com/actions/setup-python/pull/112/files#diff-7159c6139aba803e6290bd5cb287d23fR169

action.yml
```diff
@@ -8,6 +8,9 @@ inputs:
     default: '3.x'
   architecture:
     description: 'The target architecture (x86, x64) of the Python interpreter.'
+  stable:
+    description: 'Whether to download only stable versions'
```
konradpabjan, line 12: I'm not entirely sure that we need this new `stable` option.

I'm imagining a scenario that a user has the following YAML with a few versions of Python

```
   strategy
      matrix:
          python-version: [3.5.9, 3.6.10, 3.7.6, 3.8.3]
    steps:
        - uses: 'actions/setup-python@v2'
          with:
              python-version: ${{ matrix.python-version }}    
```

Now, lets say that there are a few versions of the 3.9 beta (`3.9.0-beta.4` and `3.9.0-beta.5`) that they would like to add. What I would expect the user to do is to just update the matrix. Adding `stable: false` just seems unnecessary.

```
   strategy
      matrix:
          python-version: [3.5.9, 3.6.10, 3.7.6, 3.8.3, 3.9.0-beta.4, 3.9.0-beta.5 ]
    steps:
        - uses: 'actions/setup-python@v2'
          with:
              python-version: ${{ matrix.python-version }}    
```

The fact that they have to specify `3.9.0-beta.*`  makes users aware that they are using some pre-release version, so I think making users add `stable: false` is not needed. It almost feels like an implementation detail that we don't have to expose to users.

Is there something I might be missing as to why this new option is needed?

konradpabjan: Per our offline discussions, we can remove this option 😃

`setup-go` has the same option, but we don't need it here

MaksimZhukov: Thank you! I've removed `stable` boolean input

README.md
```diff
@@ -92,6 +92,26 @@ jobs:
 
 ```
 
+Download and set up the accurate pre-release version of Python:
```
konradpabjan, line 95:
```suggestion
Download and set up a pre-release version of Python:
```

README.md
```diff
@@ -92,6 +92,26 @@ jobs:
...
+```yaml
+steps:
+- uses: actions/checkout@v2
+- uses: actions/setup-python@v2
+  with:
+    python-version: '3.9.0-beta.4'
+- run: python my_script.py
+```
+
+Download and set up the latest available version of Python (including both pre-release and stable versions):
```
konradpabjan, line 105:
```suggestion
Download and set up the latest available version of Python (includes both pre-release and stable versions):
```

README.md
```diff
@@ -92,6 +92,26 @@ jobs:
```
konradpabjan, line 92: There is a `What's new` section where support for pre-release versions can be mentioned: https://github.com/actions/setup-python#whats-new

I would suggest an extra bullet point that says: `Support for pre-release versions of Python`

MaksimZhukov: Thanks!

README.md
MaksimZhukov: Thanks!

README.md
MaksimZhukov: Thanks!

Issue #103, opened 2020-06-19
Improve error messages when installing Python on a self-hosted runner


brcrista:
In #97 for example, it's not clear why the action is failing.

Two quick improvements I think we can make:
1. When we [invoke the script to install Python](https://github.com/actions/setup-python/blob/2241d846baea6518b39bec0fb3859d75cb3d3c50/src/install-python.ts#L37), we should hook up the subprocess' `stderr` to the parent process'.
2. We already do this with `stdout`, but it only shows up in debug output.  I think we should always output this.

cc @konradpabjan

PR #113, opened 2020-07-15
Improved logging during setup

konradpabjan:
Resolves: https://github.com/actions/setup-python/issues/103

Sample run where a job was intentionally failed: https://github.com/konradpabjan/Testing2/runs/874079531

Previously, the steps in red would just not show up:

image

Some of the lines such as `Create Python 3.8.3 folder` would also only show up with debug on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,9 @@ You should specify only a major and minor version if you are okay with the most 
 
 # Using `setup-python` with a self hosted runner
 
-If you would like to use `setup-python` and a self-hosted runner, there are a few extra things you need to make sure are set up so that new versions of Python can be downloaded and configured on your runner.
+Python distributions are only available for the same [environments](https://github.com/actions/virtual-environments#available-environments) that GitHub Actions hosted environments are available for. If you are using an unsupported version of Ubuntu such as `19.04` or an alternative Linux distribution such as Fedora, `setup-python` will not work. If you have a supported `self-hosted` runner and you would like to use `setup-python`, there are a few extra things you need to make sure are set up so that new versions of Python can be downloaded and configured on your runner.
+
+If you are experience problems while configuring python on your self-hosted runner. Turn on [step debugging](https://github.com/actions/toolkit/blob/main/docs/action-debugging.md#step-debug-logs) to see addition logs.
 
 ### Windows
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -136,7 +136,9 @@ You should specify only a major and minor version if you are okay with the most
 
 # Using `setup-python` with a self hosted runner
 
-If you would like to use `setup-python` and a self-hosted runner, there are a few extra things you need to make sure are set up so that new versions of Python can be downloaded and configured on your runner.
+Python distributions are only available for the same [environments](https://github.com/actions/virtual-environments#available-environments) that GitHub Actions hosted environments are available for. If you are using an unsupported version of Ubuntu such as `19.04` or an alternative Linux distribution such as Fedora, `setup-python` will not work. If you have a supported `self-hosted` runner and you would like to use `setup-python`, there are a few extra things you need to make sure are set up so that new versions of Python can be downloaded and configured on your runner.
```
brcrista, line 139: Thanks for adding this info!

```suggestion
Python distributions are only available for the same [environments](https://github.com/actions/virtual-environments#available-environments) that GitHub Actions hosted environments are available for. If you are using an unsupported version of Ubuntu such as `19.04` or another Linux distribution such as Fedora, `setup-python` will not work. If you have a supported self-hosted runner and you would like to use `setup-python`, there are a few extra things you need to make sure are set up so that new versions of Python can be downloaded and configured on your runner.
```

brcrista: My suggestion is:
* 'alternative' -> 'another'
* Remove code formatting on 'self-hosted'

README.md
```diff
@@ -136,7 +136,9 @@ You should specify only a major and minor version if you are okay with the most
...
+
+If you are experience problems while configuring python on your `self-hosted` runner. Turn on [step debugging](https://github.com/actions/toolkit/blob/main/docs/action-debugging.md#step-debug-logs) to see addition logs.
```
brcrista, line 141:
```suggestion
If you are experience problems while configuring Python on your self-hosted runner, turn on [step debugging](https://github.com/actions/toolkit/blob/main/docs/action-debugging.md#step-debug-logs) to see addition logs.
```

dist/index.js
```diff
@@ -6355,7 +6355,10 @@ function installPython(workingDirectory) {
             silent: true,
             listeners: {
                 stdout: (data) => {
-                    core.debug(data.toString().trim());
+                    core.info(data.toString().trim());
```
brcrista, line 6358: 👍

Issue #101, opened 2020-06-08
Running on self-hosted runners leaves files in GITHUB_WORKSPACE


russellcardullo:
**Describe the bug**
The very first time setup-python is called for a specific version on a self hosted runner, the action will download files into the GITHUB_WORKSPACE and leave them there.

This causes issues for later steps, for instance if I have a step to run linters it runs on the temporary files setup-python downloads.  

**Which version of the action are you using?**

- [ ] `v1`
- [x] `v2`
- [ ] Some other tag (such as `v2.0.1` or `master`)

**Environment**
- [x] self-hosted
- [ ] Linux
- [ ] Windows
- [ ] Mac

If applicable, please specify if you're using a container

**Python Versions**
Please list all of the effected versions of Python (`3.8.2`, etc.)

**To Reproduce**

This workflow file run on a self-hosted Ubuntu 18.04 linux server will reproduce:

```
name: Test setup-python workspace
on: push
jobs:
  test:
    name: Test setup-python
    runs-on: [self-hosted, linux]
    steps:
      - uses: actions/checkout@v2
      - name: before
        run: |
          ls
          ls $GITHUB_WORKSPACE
          du -hc ./
      - name: setup python
        uses: actions/setup-python@v2
        with:
          python-version: 2.7.17
      - name: after
        run: |
          ls
          ls $GITHUB_WORKSPACE
          du -hc ./
```

After running this you should see a directory with a UUID that adds approximately ~130MB of files to the workspace.

**Run/Repo Url**
If applicable, and if your repo/run is public, please include a URL so it is easier for us to investigate.

**Screenshots**

**Additional context**

annarailton:
+1 

I've had exactly the same issue and it's caused havoc in `pytest` test discovery.

aibaars:
I think Actions should avoid writing into GITHUB_WORKSPACE as best practice, because this could cause build failures. The workspace is where the source code is checked out. The other `setup-X` actions seem to download and extract files into the toolcache. This has as added benefit that downloads are cached between runs on self-hosted runners.

jeking3:
I ran into this issue and my build started to run the python distro unit tests!

jeking3:
Per #106 reordering the checkout and setup-python resolves this.

maxim-lobanov:
We are working to provide fix for this issue

MaksimZhukov:
Hello @russellcardullo !
[...]

PR #118, opened 2020-07-17
Fix path of extracted folder for Python

nikita-bykov:
Previously, the action downloaded files into the GITHUB_WORKSPACE and leave them there. This will happen if we specify a specific location for unpacking files in method `tc.extractZip`. After removing the second parameter of this method, files will be unzipped in default directory (`/home/runner/work/_temp`).

Before changes we can see that after setup python, the size our work directory is 229MB: https://github.com/nikita-bykov/setup-python/runs/881317016?check_suite_focus=true
After changes the size is 1.1MB: https://github.com/nikita-bykov/setup-python/runs/881603419?check_suite_focus=true 

Issue: #101


src/install-python.ts
```diff
@@ -62,9 +62,9 @@ export async function installCpythonFromRelease(release: tc.IToolRelease) {
   core.info('Extract downloaded archive');
   let pythonExtractedFolder;
   if (IS_WINDOWS) {
-    pythonExtractedFolder = await tc.extractZip(pythonPath, `./${fileName}`);
```
brcrista, line 65: Looks like the `fileName` variable is unused now.

nikita-bykov: Removed this variable

PR #139, opened 2020-09-29
Add Environment Files

thboop:
The newest toolkit version uses [Environment files](https://github.com/actions/toolkit/blob/main/docs/commands.md#environment-files) to communicate with the Runner. This pr moves towards the new way of communicating with the runner and updates the appropriate toolkit version.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1245,10 +1245,72 @@ module.exports = SemVer
 
 /***/ }),
 
+/***/ 82:
+/***/ (function(__unusedmodule, exports) {
+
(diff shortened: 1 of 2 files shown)


konradpabjan:
I made the same changes except I merged them into the `releases/v1` branch (will update tags shortly).

I'm going to delete the servicing branch so that there is only `master`, `main` and `relases/v1`. Looking at telemetry, absolutely no one is using it so it's safe to delete.

Not sure when or why I originally created that branch 🤔

PR #140, opened 2020-09-30
Update README.md

Lanjewar11012001:
Improving Docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# setup-python V2
+# setup-python V 2.0
 
 <p align="left">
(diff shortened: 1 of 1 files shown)


konradpabjan:
Unfortunately I'm going to have to decline these changes.

For consistency we keep everything as `v1` or `v2` in our first party actions. If you also look at our releases there are multiple versions https://github.com/actions/setup-python/releases such as `v2.1.3` and it's constantly changing so it doesn't make sense to explicitly have `v2.0` called out. Whenever there is a new release, the `v2` tag gets updated with the latest changes.

Thank you for you willingness to make GitHub actions better!

PR #143, opened 2020-10-01
improve Comments

sushilkanathia:


hugovk (changes requested):
This is unnecessary and was possibly a Hacktoberfest spam PR. I suggest labelling `invalid` and closing.

Issue #131, opened 2020-08-14
Unable to dynamically download and install manifests - library not found/path issue


magnetikonline:
**Describe the bug**

**Which version of the action are you using?**

- [ ] `v1`
- [x] `v2`
- [ ] Some other tag (such as `v2.0.1` or `master`)

**Environment**
- [x] self-hosted
- [x] Linux
- [ ] Windows
- [ ] Mac

Our self hosted runners are using a Ubuntu 18.04 base distro image.

**Python Versions**
Please list all of the effected versions of Python (`3.8.2`, etc.)
- `3.7.5`
- But really, any version which is not installed already to our self hosted runners.

**To Reproduce**
Setup workflow with the following run action:

```yaml
- name: Test Setup Python
   uses: actions/setup-python@v2
   with:
      python-version: '3.7.5'
```

The error received is:
```
Create Python 3.7.5 folder
Copy Python binaries to hostedtoolcache folder
Create additional symlinks (Required for the UsePythonVersion Azure Pipelines task and the setup-python GitHub Action)
Upgrading PIP...
##[error]./python: error while loading shared libraries: libpython3.7m.so.1.0: cannot open shared object file: No such file or directory
##[error]The process '/bin/bash' failed with exit code 127
```

This is the same deal with other editions, during install the Python library can't be located - I'm not sure if the issue lies with the Action itself, or the packaged Python installers from https://github.com/actions/python-versions that need modification.

I have been pre-loading specific Python versions to our self hosted runner - and in those cases I've been able to set `LD_LIBRARY_PATH` to get installs to take - not sure if this needs to be done by default for all manifests? Example, with a Python manifest downloaded and unzipped:

```sh
export LD_LIBRARY_PATH="/home/runner/_work/_tool/Python/3.7.5/x64/lib"
bash ./setup.sh
```

This will successfully install.

So I wonder, is the solution to append the installed library path to `LD_LIBRARY_PATH` in the context of `./setup.sh` to provide a hint to `pip` where to locate required Python libs?

E.g. here: https://github.com/actions/python-versions/blob/ca3edf54460c3581d7285dd0fba82fa7d9215b6b/installers/nix-setup-template.sh#L51

**Screenshots**

![Screen Shot 2020-08-14 at 10 40 01 pm](https://user-images.githubusercontent.com/1818757/90250247-24553c00-de7f-11ea-8fea-57471067d4c5.png)

**Additional context**
Feels somewhat related to #115

rmileskcl:
[...]

PR #144, opened 2020-10-01
Inject LD_LIBRARY_PATH library path into Python manifest install and setup

magnetikonline:
This PR sets up the env var `LD_LIBRARY_PATH` for use when installing a fresh Python manifest and configuring it for use - ensures the Python manifest version installed can have it's libs correctly located.

Note this _only_ applies to action runs under Linux based OSes.

While in the area, noted small typo in `docs/contributors.md`.

Related prior discussions:

- https://github.com/actions/setup-python/issues/131
- https://github.com/actions/python-versions/pull/49

### DIFF
--- a/src/find-python.ts
+++ b/src/find-python.ts
@@ -9,6 +9,7 @@ import * as core from '@actions/core';
 import * as tc from '@actions/tool-cache';
 
 const IS_WINDOWS = process.platform === 'win32';
+const IS_LINUX = process.platform === 'linux';
 
 // Python has "scripts" or "bin" directories where command-line tools that come with packages are installed.
 // This is where pip is, along with anything that pip installs.
@@ -109,6 +110,18 @@ async function useCpythonVersion(
   }
 
   core.exportVariable('pythonLocation', installDir);
+
+  if (IS_LINUX) {
+    const libPath = process.env.LD_LIBRARY_PATH
+      ? `:${process.env.LD_LIBRARY_PATH}`
+      : '';
+    const pyLibPath = path.join(installDir, 'lib');
+
+    if (!libPath.split(':').includes(pyLibPath)) {
+      core.exportVariable('LD_LIBRARY_PATH', pyLibPath + libPath);
+    }
+  }
+
   core.addPath(installDir);
   core.addPath(binDir(installDir));
 
--- a/src/install-python.ts
+++ b/src/install-python.ts
@@ -13,6 +13,7 @@ const MANIFEST_REPO_BRANCH = 'main';
 export const MANIFEST_URL = `https://raw.githubusercontent.com/${MANIFEST_REPO_OWNER}/${MANIFEST_REPO_NAME}/${MANIFEST_REPO_BRANCH}/versions-manifest.json`;
 
 const IS_WINDOWS = process.platform === 'win32';
+const IS_LINUX = process.platform === 'linux';
 
 export async function findReleaseFromManifest(
   semanticVersionSpec: string,
@@ -35,6 +36,10 @@ export async function findReleaseFromManifest(
 async function installPython(workingDirectory: string) {
   const options: ExecOptions = {
     cwd: workingDirectory,
+    env: {
+      ...process.env,
+      ...(IS_LINUX && {LD_LIBRARY_PATH: path.join(workingDirectory, 'lib')})
+    },
     silent: true,
     listeners: {
       stdout: (data: Buffer) => {
--- a/dist/index.js
+++ b/dist/index.js
@@ -6422,6 +6422,7 @@ var __importStar = (this && this.__importStar) || function (mod) {
     return result;
 };
 Object.defineProperty(exports, "__esModule", { value: true });
+const path = __importStar(__webpack_require__(622));
 const core = __importStar(__webpack_require__(470));
 const tc = __importStar(__webpack_require__(533));
 const exec = __importStar(__webpack_require__(986));
@@ -6432,6 +6433,7 @@ const MANIFEST_REPO_NAME = 'python-versions';
 const MANIFEST_REPO_BRANCH = 'main';
 exports.MANIFEST_URL = `https://raw.githubusercontent.com/${MANIFEST_REPO_OWNER}/${MANIFEST_REPO_NAME}/${MANIFEST_REPO_BRANCH}/versions-manifest.json`;
 const IS_WINDOWS = process.platform === 'win32';
+const IS_LINUX = process.platform === 'linux';
 function findReleaseFromManifest(semanticVersionSpec, architecture) {
     return __awaiter(this, void 0, void 0, function* () {
         const manifest = yield tc.getManifestFromRepo(MANIFEST_REPO_OWNER, MANIFEST_REPO_NAME, AUTH, MANIFEST_REPO_BRANCH);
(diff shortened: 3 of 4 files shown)


magnetikonline:
Here we go @maxim-lobanov / @MaksimZhukov - this does the job nicely.

src/find-python.ts
```diff
@@ -109,6 +110,18 @@ async function useCpythonVersion(
...
 
   core.exportVariable('pythonLocation', installDir);
+
+  if (IS_LINUX) {
+    const libPath = process.env.LD_LIBRARY_PATH
+      ? `:${process.env.LD_LIBRARY_PATH}`
+      : '';
+    const pyLibPath = path.join(installDir, 'lib');
+
+    if (!libPath.split(':').includes(pyLibPath)) {
```
maxim-lobanov, line 120: I could be wrong but I think it is better to always add `pyLibPath` to begin of the LD_LIBRARY_PATH because if `pyLibPath` in LD_LIBRARY_PATH but it is not the first one, it can cause incorrect work of Python.

magnetikonline: @maxim-lobanov the libs loaded (filename) are unique to the Python version. Yeah I was trying to avoid the situation if someone did decide to include the `setup-python` action multiple times for the same version - and duplication ends up being a problem?

Maybe it's better to:

- Split existing path.
- If found - remove item.
- Add item to start of list.
- `join()` list - as the new mutation of `LD_LIBRARY_PATH`?

What do you think @maxim-lobanov ?

src/install-python.ts
```diff
@@ -35,6 +36,10 @@ export async function findReleaseFromManifest(
 async function installPython(workingDirectory: string) {
   const options: ExecOptions = {
     cwd: workingDirectory,
+    env: {
+      ...process.env,
+      ...(IS_LINUX && {LD_LIBRARY_PATH: path.join(workingDirectory, 'lib')})
```
maxim-lobanov, line 41: As I understand, we set `env` here to make sure that [setup.sh](https://github.com/actions/python-versions/blob/main/installers/nix-setup-template.sh#L52) script will run pip commands successfully.

But please pay attention how it works:
- `setup.sh` copies the content of `workingDirectory` to `$AGENT_TOOLSDIRECTORY\Python\<version>\x64`
- run `python -m ensurepip` in `$AGENT_TOOLSDIRECTORY\Python\<version>\x64`

So ideally, we should set `LD_LIBRARY_PATH` as `$AGENT_TOOLSDIRECTORY\Python\<version>\x64` but you set it to `workingDirectory`. 

Probably, it is not a problem at all. Just want to bring it up

magnetikonline: Yeah I didn't see this one as a problem - as `setup.sh` is always at the root of the manifest and `lib/` is always inline as a directory from that archive - so we can _pretty safely_ assume that if `workingDirectory` is the right place to call `setup.sh` - we know `lib/` is also here.

src/find-python.ts
maxim-lobanov: Even someone decide to include `setup-python` action multiple times for the same version (very strange case :) ), duplication won't cause problems.

If we would like to avoid it, your current approach should work (since `the libs loaded (filename) are unique to the Python version`).

cc: @konradpabjan

src/install-python.ts
maxim-lobanov: Sounds good! Let's keep it this way

src/find-python.ts
magnetikonline: Cheers - yeah let's see what @konradpabjan thinks as a second opinion.

Able to pivot this to above if needed (re-insert as the first component).

maxim-lobanov (approved):

magnetikonline:
Thanks @maxim-lobanov 👍 

I had a bit more of a think about this - if we _do_ always want the current Python lib path first - and never duplicated, this would be the implement. Nothing too tricky, should we swap this out?

```js
core.exportVariable('pythonLocation', installDir);

if (IS_LINUX) {
  const libPaths = (process.env.LD_LIBRARY_PATH || '').split(':');
  const pyLibPath = path.join(installDir, 'lib');

  const foundIdx = libPaths.indexOf(pyLibPath);
  if (foundIdx > -1) {
    libPaths.splice(foundIdx, 1);
  }

  libPaths.unshift(pyLibPath);
  core.exportVariable('LD_LIBRARY_PATH', libPaths.join(':'));
}

core.addPath(installDir);
core.addPath(binDir(installDir));
```

src/find-python.ts
konradpabjan: @magnetikonline

Overall the current approach seems good 👍 

I do have once concern though around where where the env variable is set:
`core.exportVariable('LD_LIBRARY_PATH', pyLibPath + libPath);` 

We're currently omitting a `:` between the values (this works if libpath is empty which I think is the case with our hosted environments). When this value is not empty, things could break.

I think `core.exportVariable('LD_LIBRARY_PATH', pyLibPath + ':' +  libPath);` or even `core.exportVariable(pylibpath.concat(':', libPath);` is needed.

The order of `pyLibPath ` and `libPath` doesn't really make a difference since they're unique. No preference with regards to that.

maxim-lobanov:
@magnetikonline , have you tested that your changes work as expected for you?
You can call your fork in workflow just for testing purpose
```
- uses: magnetikonline/setup-python@inject-ld_library_path
  with:
    python-version: '3.x'
```

src/find-python.ts
magnetikonline: @konradpabjan this is covered (the `:`):

```js
const libPath = process.env.LD_LIBRARY_PATH
      ? `:${process.env.LD_LIBRARY_PATH}`
      : '';
```

Note if `process.env.LD_LIBRARY_PATH` does exist `libPath` is set to `:` _plus_ the current library path. Make sense?

magnetikonline:
> @magnetikonline , have you tested that your changes work as expected for you?
> You can call your fork in workflow just for testing purpose

I have @maxim-lobanov  - tested last week. 👍

src/find-python.ts
konradpabjan: Got it! Thanks for the clarification. I accidently missed it earlier 😞

With everything looks great!

magnetikonline: Not a problem - it's an easy miss 😄

konradpabjan (approved):
Thanks for the PR @magnetikonline 🎉 

I'm going to go ahead and approve this and merge it into the `main` branch. I don't expect any problems but this will allow us to test on a large subset of users that pin to `@main`. If everything is good, I'll create a new release and update the `v2` tag in a few days.

magnetikonline:
Awesome - thanks for your time @konradpabjan / @maxim-lobanov - I just retested `setup-python@main` vs. `setup-python@v2` and yes - it's now 100% when using `main`.

Will be great to see this in V2 soon! 👍

magnetikonline:
Hello @konradpabjan - just wondering if we have an ETA for tagging `v2` with this PR/commit?

Thanks!

konradpabjan:
@magnetikonline I've created a new release (`v2.1.4`) and I've updated the `v2` tag with the latest changes.

Thanks again for the PR!

magnetikonline:
Fantastic - thanks @konradpabjan 👍

PR #147, opened 2020-10-02
Update README.md

Educatemeans:
Improve docs


konradpabjan:
Thanks for the PR!

Unfortunately I'm going to have to decline these changes. Grammatically I think it sounds worse and `that they` is not needed

PR #164, opened 2020-12-04
Update "pypy3" keyword to point to 3.6 for backward compatibility

maxim-lobanov:
Currently, PyPy 2.7 and 3.6 are cached on Hosted images and `3.x` selector works great.
In future, we are planning to add PyPy 3.7 too. In current implementation, selector 3.x will consume 3.7 and all customers will be switched automatically to 3.7 instead of 3.6 that could cause breaking changes.

We would like to make selector more specific to choose exact `3.6` to make sure that adding 3.7 to images won't break customers.
We are going to implement PyPy 3.7 support soon so customers will be able to select 3.6 and 3.7 separately.

**Note:**
Please pay attention that after this PR, `core.setOutput('python-version', impl);` will be set to `pypy3.6` instead of `pypy3`. Not sure about impact from this change

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,3 +90,24 @@ jobs:
 
     - name: Run simple code
       run: python -c 'import math; print(math.factorial(5))'
+
+  setup-pypy:
+    name: Setup PyPy ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: setup-python pypy3
+      uses: ./
+      with:
+        python-version: 'pypy3'
+
+    - name: setup-python pypy2
+      uses: ./
+      with:
+        python-version: 'pypy2'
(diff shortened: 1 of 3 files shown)


konradpabjan (approved):
Looks good to me! 🚀 

Note: The `v2` tag won't be updated with these changes until next week after Universe, due to code freeze

EwoutH:
PyPy 3.6 is currently not a stable version anymore. Should the `pypy3` tag be updated to point to PyPy3.8, or does that break backwards compatibility?

I think ideally the `pypy3` tag should point to the latest PyPy version installed on [default images](https://github.com/actions/virtual-environments), so [currently](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#pypy) PyPy3.8.

nulano:
> PyPy 3.6 is currently not a stable version anymore. Should the `pypy3` tag be updated to point to PyPy3.8, or does that break backwards compatibility?
> 
> I think ideally the `pypy3` tag should point to the latest PyPy version installed on [default images](https://github.com/actions/virtual-environments), so [currently](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#pypy) PyPy3.8.

Related: https://github.com/actions/setup-python/issues/244

EwoutH:
Is it correct that there isn't a keyword for the latest stable PyPy version available right now? I think that some keyword to use the latest PyPy version (like `pypy3` or `pypy-3.x`) would be useful.

Issue #17, opened 2019-08-28
Add pypy-nightly and pypy3-nightly


Julian:
Besides #16 for the latest releases, it would be nice to be able to build against the nightly versions of PyPy as well: http://buildbot.pypy.org/nightly/trunk/

```

pypy-c-jit-latest-linux64.tar.bz2 | 25M | 2019-08-28 |   |  
-- | -- | -- | -- | --
pypy-c-jit-latest-linux.tar.bz2 | 24M | 2019-08-28 |   |  
pypy-c-jit-latest-aarch64.tar.bz2 | 23M | 2019-08-28 |   |  
pypy-c-jit-latest-win32.zip | 27M | 2019-08-28 |   |  
pypy-c-jit-latest-s390x.tar.bz2 | 25M | 2019-08-28 |   |  
pypy-c-jit-latest-linux-armhf-raspbian.tar.bz2 | 23M | 2018-12-01 |   |  
pypy-c-jit-latest-linux-armel.tar.bz2 | 23M | 2018-11-21 |   |  
pypy-c-jit-latest-osx64.tar.bz2 | 18M | 2019-08-28 |
```

konradpabjan:
Support for this has been added with the latest release of `setup-python@v2` 

For more information, checkout: https://github.com/actions/setup-python#specifying-a-pypy-version

Julian:
Amazing, thanks!

PR #168, opened 2020-12-14
Adding support for more PyPy versions and installing them on-flight

dmitry-shibanov:
## Description

This pull-request adds support for specifying PyPy via PyPy version and Python version.
Closes https://github.com/actions/setup-python/issues/136 and https://github.com/actions/setup-python/issues/17
### Supported input syntax
The new supported syntax:
```
pypy-3.6-v7.3.3
pypy-3.6-v7.3.x
pypy-3.6-v7.x
pypy-3.6 (pypy-3.7, pypy-2.7) (points to latest version of PyPy)
pypy-3.6-v7.3.3rc1
pypy-3.6-nightly
```
Also, we keep old syntax for backward compatibility:
```
pypy2 # (points to 2.7)
pypy3 # (points to 3.6)
```
We don't have plans to deprecate and we are going to keep them for now.

#### The list of available versions
Current implementation uses the list of available versions from https://downloads.python.org/pypy/versions.json 
It was discussed and confirmed with PyPy team: https://foss.heptapod.net/pypy/pypy/-/issues/3354

#### Storing exact version of PyPy in "PYPY_VERSION" file in PyPy folder
Currently, PyPy versions are saved in toolcache. Toolcache has the following structure `<toolcache_root>/PyPy/<python_version>/x64`.
So it is not possible to determine from path what version of PyPy is located in folder. For example, on machine, we have PyPy 7.3.1 with python 3.6 in toolcache on path `<toolcache_root>/PyPy/3.6/x64`. We need some way to determine what version of PyPy in this folder.
We have consider calling `pypy --version` and parsing output but this approach doesn't work for RC versions because `pypy --version` doesn't show exact RC number in output.
We have decided to write exact installed version in file and put it to PyPy folder. This approach works perfectly and it is faster than parsing output. These files will be added to Hosted images in a few weeks.
The single drawback, action won't consume PyPy that was installed manually without this action.

#### Installation time
- MacOS: 20 seconds
- Ubuntu: 13 seconds
- Windows: 40-55 seconds
- All versions of Python for the latest PyPy will be cached on image and updated with a week delay.

#### Support of RC and nightly versions
- The RC versions and nightly versions are also supported.
- Nightly versions are not saved in local cache and always downloaded and installed for now

#### Self-hosted agents
This action also supports self-hosted agents and works on them

### DIFF
--- a/.github/workflows/test-pypy.yml
+++ b/.github/workflows/test-pypy.yml
@@ -9,7 +9,7 @@ on:
     paths-ignore:
       - '**.md'
   schedule:
-    - cron: 0 0 * * *
+    - cron: 30 3 * * *
 
 jobs:
   setup-pypy:
--- a/.github/workflows/test-pypy.yml
+++ b/.github/workflows/test-pypy.yml
@@ -1,0 +1,47 @@
+name: Validate PyPy e2e
+on: 
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+  schedule:
+    - cron: 0 0 * * *
+
+jobs:
+  setup-pypy:
+    name: Setup PyPy ${{ matrix.pypy }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-18.04, ubuntu-20.04]
+        pypy:
+        - 'pypy-2.7'
+        - 'pypy-3.6'
+        - 'pypy-3.7'
+        - 'pypy-2.7-v7.3.2'
+        - 'pypy-3.6-v7.3.2'
+        - 'pypy-3.7-v7.3.2'
+        - 'pypy-3.6-v7.3.x'
+        - 'pypy-3.7-v7.x'
+        - 'pypy-3.6-v7.3.3rc1'
+        - 'pypy-3.7-nightly'
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+  
+      - name: setup-python ${{ matrix.pypy }}
+        uses: ./
+        with:
+          python-version: ${{ matrix.pypy }}
+  
+      - name: PyPy and Python version
+        run: python --version
+  
+      - name: Run simple code
+        run: python -c 'import math; print(math.factorial(5))'
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -1,4 +1,4 @@
-name: Validate 'setup-python'
+name: Validate Python e2e
 on: 
   push:
     branches:
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04]
+        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04]
+        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
         python: [3.5.4, 3.6.7, 3.7.5, 3.8.1]
     steps:
     - name: Checkout
@@ -68,7 +68,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04]
+        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
@@ -91,13 +91,13 @@ jobs:
     - name: Run simple code
       run: python -c 'import math; print(math.factorial(5))'
 
-  setup-pypy:
+  setup-pypy-legacy-way:
     name: Setup PyPy ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04]
+        os: [macos-latest, windows-latest, ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -9,7 +9,7 @@ on:
     paths-ignore:
       - '**.md'
   schedule:
-    - cron: 0 0 * * *
+    - cron: 30 3 * * *
 
 jobs:
   default-version:
@@ -91,7 +91,7 @@ jobs:
     - name: Run simple code
       run: python -c 'import math; print(math.factorial(5))'
 
-  setup-pypy-legacy-way:
+  setup-pypy-legacy:
     name: Setup PyPy ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ This action sets up a Python environment for use in actions by:
   - Allows for pinning to a specific patch version of Python without the worry of it ever being removed or changed.
 - Automatic setup and download of Python packages if using a self-hosted runner.
 - Support for pre-release versions of Python.
+- Support for installation any version of PyPy on-flight
 
 # Usage
 
@@ -40,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '2.x', '3.x', 'pypy2', 'pypy3' ]
+        python-version: [ '2.x', '3.x', 'pypy-2.7', 'pypy-3.6', 'pypy-3.7' ]
     name: Python ${{ matrix.python-version }} sample
     steps:
       - uses: actions/checkout@v2
@@ -60,7 +61,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [2.7, 3.6, 3.7, 3.8, pypy2, pypy3]
+        python-version: [2.7, 3.6, 3.7, 3.8, pypy-2.7, pypy-3.6]
         exclude:
           - os: macos-latest
             python-version: 3.8
@@ -91,7 +92,6 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - run: python my_script.py
-
 ```
 
 Download and set up an accurate pre-release version of Python:
@@ -114,6 +114,27 @@ steps:
 - run: python my_script.py
 ```
 
+Download and set up PyPy:  
+
+```yaml
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version:
+        - pypy-3.6 # the latest available version of PyPy
+        - pypy-3.7 # the latest available version of PyPy
+        - pypy-3.7-v7.3.3 # Python 3.7 and PyPy 7.3.3
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - run: python my_script.py
+```
+More details on PyPy syntax and examples of using preview / nightly versions of PyPy can be found in [Available versions of PyPy](#available-versions-of-pypy) section.
+
 # Getting started with Python + Actions
 
 Check out our detailed guide on using [Python with GitHub Actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-python-with-github-actions).
@@ -129,7 +150,21 @@ Check out our detailed guide on using [Python with GitHub Actions](https://help.
     - If the exact patch version doesn't matter to you, specifying just the major and minor version will get you the latest preinstalled patch version. In the previous example, the version spec `3.8` will use the `3.8.2` Python version found in the cache.
 - Downloadable Python versions from GitHub Releases ([actions/python-versions](https://github.com/actions/python-versions/releases)).
     - All available versions are listed in the [version-manifest.json](https://github.com/actions/python-versions/blob/main/versions-manifest.json) file.
-    - If there is a specific version of Python that is not available, you can open an issue here.
+    - If there is a specific version of Python that is not available, you can open an issue here
+    
+ # Available versions of PyPy
+ 
+ `setup-python` is able to configure PyPy from two sources:
+ 
+- Preinstalled versions of PyPy in the tools cache on GitHub-hosted runners
+  - For detailed information regarding the available versions of PyPy that are installed see [Supported software](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-software).
+  - For the latest PyPy release, all version of Python are cached.
+  - Cache is updated with 1-2 weeks delay. if you specify PyPy as `pypy-3.6`, the version from cache will be used although a new version is available. If you need to start using the recently released version right after release, you should specify exact PyPy version `pypy-3.6-v7.3.3`.
+
+- Downloadable PyPy versions from [official PyPy site](https://downloads.python.org/pypy/).
+  - All available versions are listed in the [versions.json](https://downloads.python.org/pypy/versions.json) file.
+  - PyPy < 7.3.3 are not available to install on-flight.
+  - If some versions are not available, you can open an issue in https://foss.heptapod.net/pypy/pypy/
 
 # Hosted Tool Cache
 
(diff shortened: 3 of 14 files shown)


src/install-pypy.ts
```diff
@@ -0,0 +1,216 @@
...
+    true
+  );
+}
+
+async function installPip(pythonLocation: string) {
+  core.info('Installing and updating pip');
+  const pythonBinary = path.join(pythonLocation, 'python');
+  await exec.exec(`${pythonBinary} -m ensurepip`);
+  // TO-DO should we skip updating of pip ?
+  await exec.exec(
```
maxim-lobanov, line 115: Question for discussion:
Do we really need to update pip after installing? I agree that `-m ensurepip` is necessary but I wonder about ` -m pip install --ignore-installed pip`.
Is it a problem if pip is a bit outdated?

dmitry-shibanov: I think we should reinstall/update pip because if we skip upgrade, it can raise an error in the further steps. It won’t be able to find some packages. For example, I've got an error with arrow package.  Moreover, the documentation says that it is better to upgrade pip after the [ensurepip](https://pip.pypa.io/en/stable/installing/#using-ensurepip). Besides, we have the [issue](https://github.com/actions/virtual-environments/issues/144) in virtual environment about updating pip.

maxim-lobanov: Sounds good, thank you for explanation. I worried that it takes additional 5-10 seconds but looks like it makes sense.

src/setup-python.ts
```diff
@@ -1,15 +1,29 @@
...
+}
+
 async function run() {
   try {
     let version = core.getInput('python-version');
     if (version) {
       const arch: string = core.getInput('architecture') || os.arch();
-      const installed = await finder.findPythonVersion(version, arch);
-      core.info(`Successfully setup ${installed.impl} (${installed.version})`);
+      if (isPyPyVersion(version)) {
```
konradpabjan, line 16: Question, how does back-compat work here since our existing pypy2 and pypy3 inputs would return false for `isPyPyVersion()`?

src/find-pypy.ts
```diff
@@ -0,0 +1,116 @@
...
+  return {installDir, resolvedPythonVersion, resolvedPyPyVersion};
+}
+
+function parsePyPyVersion(versionSpec: string): IPyPyVersionSpec {
+  const versions = versionSpec.split('-').filter(item => !!item);
+
+  if (versions.length < 2) {
+    throw new Error(
+      "Invalid 'version' property for PyPy. PyPy version should be specified as 'pypy-<python-version>'. See readme for more examples."
+    );
```
konradpabjan, line 98: I would change this to use [core.setError](https://github.com/actions/toolkit/tree/main/packages/core#exit-codes), the process will still fail and exit so the behavior will be the same

Nit, I would also change the ending from `See readme for more examples.` to `See README for examples and documentation.`

src/find-pypy.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+  const pythonVersion = versions[1];
```
konradpabjan, line 100: I would add some extra validation here, just to check that 3.6 is actually a number. Users might put in a typo or something so it would be safe to error out here with a clear message

src/install-pypy.ts
```diff
@@ -0,0 +1,216 @@
...
+export function getPyPyBinaryPath(installDir: string) {
+  const _binDir = path.join(installDir, 'bin');
+  return IS_WINDOWS ? installDir : _binDir;
+}
+
+function isNightlyKeyword(pypyVersion: string) {
+  return pypyVersion === 'nightly';
+}
+
+export function pypyVersionToSemantic(versionSpec: string) {
```
konradpabjan, line 213: I _think_ this looks good! 😄 I'm not a regex expert but some tests for this in particular would go a long way!

src/install-pypy.ts
```diff
@@ -0,0 +1,216 @@
...
+    throw new Error(
+      `PyPy version ${pythonVersion} (${pypyVersion}) with arch ${architecture} not found`
+    );
+  }
+
+  const {foundAsset, resolvedPythonVersion, resolvedPyPyVersion} = releaseData;
+  let downloadUrl = `${foundAsset.download_url}`;
+
+  core.info(`Downloading PyPy from "${downloadUrl}" ...`);
+  const pypyPath = await tc.downloadTool(downloadUrl);
```
konradpabjan, line 38: This is good! 👍 The `tc.downloadTool` function has built-in retry capabilities so if something goes wrong no need to worry about anything. 

Extraction also happens to a temp folder that should get cleaned up.

src/install-pypy.ts
```diff
@@ -0,0 +1,216 @@
...
+}
+
+function findRelease(
+  releases: IPyPyManifestRelease[],
+  pythonVersion: string,
+  pypyVersion: string,
+  architecture: string
+) {
+  const filterReleases = releases.filter(item => {
+    const isPythonVersionSatisfies = semver.satisfies(
```
konradpabjan, line 127: Nit Pick: I would change `isPythonVersionSatisfies` to `isPythonVersionSatisfied`, same thing with `isPyPyVersionSatisfied` below

`isArchExists` also sounds a little incorrect. I would change that to `isArchPresent` as it's like saying "Is the architecture present?". `isArchExists` would make sense if we were saying "Does the architecture exist?" but since we want to prefix with "is", the former is the better naming convention.

src/install-pypy.ts
konradpabjan: I also think we should reinstall/update pip, it's currently the default behavior that we have.

For a future discussion, it might be worthwhile to have an optional parameter that allows the user to opt-out of upgrading pip, however, for the current scenario we should keep it as it.

konradpabjan (commented):
Awesome job so far! Just left a few minor comments in a few places.

src/find-pypy.ts
```diff
@@ -0,0 +1,125 @@
...
+      architecture
+    ));
+  }
+
+  const pipDir = IS_WINDOWS ? 'Scripts' : 'bin';
+  const _binDir = path.join(installDir, pipDir);
+  const pythonLocation = pypyInstall.getPyPyBinaryPath(installDir);
+  core.exportVariable('pythonLocation', pythonLocation);
+  core.addPath(pythonLocation);
+  core.addPath(_binDir);
```
maxim-lobanov, line 52: Should we also do `setOutput`?
https://github.com/actions/setup-python/blob/main/src/find-python.ts#L70

src/find-pypy.ts
```diff
@@ -0,0 +1,125 @@
...
+    process.exit();
+  }
+
+  return {
+    pypyVersion: pypyVersion,
+    pythonVersion: pythonVersion
+  };
+}
+
+function getPyPyVersionFromPath(installDir: string) {
```
maxim-lobanov, line 123: Should we move it to utils?

src/setup-python.ts
maxim-lobanov: `pypy2` and `pypy3` still use the previous logic and usePyPy function from CPython.
We didn't touch that logic at all in this PR: https://github.com/dmitry-shibanov/setup-python/blob/v-dmshib/add-pypy-installation/src/find-python.ts#L38

src/install-pypy.ts
dmitry-shibanov: Resolved.

src/find-pypy.ts
dmitry-shibanov: Added additional validation.

src/find-pypy.ts
dmitry-shibanov: I prefer to keep throw exception here and this exception will be handled on top level: https://github.com/actions/setup-python/blob/3b3f2de1b1f7c1270fc5b9c37a904fa785e9ae94/src/setup-python.ts#L7. Core setFailed does not exit program and requires additional process exit. Exception looks clear way.

.github/workflows/test-pypy.yml
```diff
@@ -0,0 +1,47 @@
...
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+  schedule:
+    - cron: 0 0 * * *
```
konradpabjan, line 12: Nit, I would change this to `30 3 * * *`, we get bombarded with a wave of scheduled workflows at midnight UTC because most users default to `0 0 * * *`. When can help not contribute to that load by running at a slightly different time 😃

.github/workflows/test-python.yml
```diff
@@ -91,13 +91,13 @@ jobs:
     - name: Run simple code
       run: python -c 'import math; print(math.factorial(5))'
 
-  setup-pypy:
+  setup-pypy-legacy-way:
```
konradpabjan, line 94: Nit: I would change this to just `setup-pypy-legacy`

README.md
```diff
@@ -17,6 +17,7 @@ This action sets up a Python environment for use in actions by:
   - Allows for pinning to a specific patch version of Python without the worry of it ever being removed or changed.
 - Automatic setup and download of Python packages if using a self-hosted runner.
 - Support for pre-release versions of Python.
+- Support for installation any version of PyPy on-flight
```
konradpabjan, line 20: I would change this to `Support for installing any version of PyPy on-flight`

README.md
```diff
@@ -114,6 +114,27 @@ steps:
...
+        - pypy-3.7 # the latest available version of PyPy
+        - pypy-3.7-v7.3.3 # Python 3.7 and PyPy 7.3.3
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - run: python my_script.py
+```
+More details on PyPy syntax and examples of using preview / nightly versions of PyPy can be found in [Available versions of PyPy](#available-versions-of-pypy) section.
```
konradpabjan, line 136: Nit: versions of PyPy can be found in **the** [Available versions

README.md
```diff
@@ -129,7 +150,21 @@ Check out our detailed guide on using [Python with GitHub Actions](https://help.
...
-    - If there is a specific version of Python that is not available, you can open an issue here.
+    - If there is a specific version of Python that is not available, you can open an issue here
+    
+ # Available versions of PyPy
+ 
+ `setup-python` is able to configure PyPy from two sources:
+ 
+- Preinstalled versions of PyPy in the tools cache on GitHub-hosted runners
+  - For detailed information regarding the available versions of PyPy that are installed see [Supported software](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-software).
+  - For the latest PyPy release, all version of Python are cached.
```
konradpabjan, line 161: all version**s** of Python are cached.

README.md
```diff
@@ -129,7 +150,21 @@ Check out our detailed guide on using [Python with GitHub Actions](https://help.
...
+  - Cache is updated with 1-2 weeks delay. if you specify PyPy as `pypy-3.6`, the version from cache will be used although a new version is available. If you need to start using the recently released version right after release, you should specify exact PyPy version `pypy-3.6-v7.3.3`.
```
konradpabjan, line 162: A few changes I would make: 

Cache is updated with **a** 1-2 **week** delay. **I**f you specify the PyPy version as `pypy-3.6`, **the cached version will be used** although a new**er** version is available. If you need to start using the recently released version right after release, you should specify **the** exact PyPy version **using** `pypy-3.6-v7.3.3`.

README.md
```diff
@@ -129,7 +150,21 @@ Check out our detailed guide on using [Python with GitHub Actions](https://help.
...
+
+- Downloadable PyPy versions from [official PyPy site](https://downloads.python.org/pypy/).
+  - All available versions are listed in the [versions.json](https://downloads.python.org/pypy/versions.json) file.
```
konradpabjan, line 165: Downloadable PyPy versions from **the** [official PyPy site]
All available versions **that we can download** are listed in...

README.md
```diff
@@ -155,6 +190,20 @@ You should specify only a major and minor version if you are okay with the most
   - There will be a single patch version already installed on each runner for every minor version of Python that is supported.
   - The patch version that will be preinstalled, will generally be the latest and every time there is a new patch released, the older version that is preinstalled will be replaced.
   - Using the most recent patch version will result in a very quick setup since no downloads will be required since a locally installed version Python on the runner will be used.
+  
+# Specifying a PyPy version
+The version of PyPy should be specified in the format `pypy-<python_version>[-v<pypy_version>]`.  
+Parameter `<pypy_version>` is optional and can be skipped. The latest version will be used in this case.
```
konradpabjan, line 196: I would change this to: `The <pypy_version> parameter is optional and can be skipped.`

README.md
```diff
@@ -155,6 +190,20 @@ You should specify only a major and minor version if you are okay with the most
...
+
+```
+pypy-3.6 # the latest available version of PyPy
+pypy-3.7 # the latest available version of PyPy
+pypy-2.7 # the latest available version of PyPy
```
konradpabjan, line 201: I would change the comment for each of these to say something like `the latest available version of PyPy that supports Python 3.6` etc.

There is also the example `Download and set up PyPy` YAML above that has similar comments, I would change that as well.

konradpabjan (commented):
Looking good! Leaving just a few minor grammar fixes

maxim-lobanov:
@mattip , @jarrodmillman we are on the final line with these changes and are going to merge it in 1-2 days. 
Please let us know if you see any issues with this implementation and have any suggestions

mattip:
Thanks, the design looks good to me.

mattip:
Maybe the cibuildwheel people have some comments? ping @joerick

joerick:
cibuildwheel is CI-agnostic so we probably wouldn't use this, but I had a quick read through and it looks good to me! @yannickjadoul was our point guy on PyPy, he might have some input.

mattip:
@joerick: right, sorry I should have been more specific. Your community have experience with specifying and caching, so I thought maybe your well-honed instincts would be of benefit in reviewing this kind of version-specifier design.

src/find-pypy.ts
dmitry-shibanov: Discussed offline

src/find-pypy.ts
dmitry-shibanov: Yes, sure. Moved it.

README.md
dmitry-shibanov: Thank you. Fixed.

README.md
dmitry-shibanov: Thank you. Fixed.

README.md
dmitry-shibanov: Thank you. Fixed.

README.md
dmitry-shibanov: Thank you. Fixed.

README.md
dmitry-shibanov: Thank you. Fixed.

README.md
dmitry-shibanov: Thank you. Fixed.

.github/workflows/test-pypy.yml
dmitry-shibanov: Thank you. Fixed.

YannickJadoul:
> cibuildwheel is CI-agnostic so we probably wouldn't use this, but I had a quick read through and it looks good to me! @YannickJadoul was our point guy on PyPy, he might have some input.

I've mostly been liaising with @mattip and the other developers for `cibuildwheel`, though, so I'm not sure I have a lot extra to add here :-)

One thing that we also still haven't figured out in `cibuildwheel` is how many different ABI versions of PyPy to offer, and when to consider a PyPy version really outdated? But this PR seems to just be offering them all, as it serves a slightly different purpose to `cibuildwheel`.
https://downloads.python.org/pypy/versions.json shows less versions than https://downloads.python.org/pypy/, but maybe that's on purpose?

maxim-lobanov:
> > cibuildwheel is CI-agnostic so we probably wouldn't use this, but I had a quick read through and it looks good to me! @YannickJadoul was our point guy on PyPy, he might have some input.
> 
> I've mostly been liaising with @mattip and the other developers for `cibuildwheel`, though, so I'm not sure I have a lot extra to add here :-)
> 
> One thing that we also still haven't figured out in `cibuildwheel` is how many different ABI versions of PyPy to offer, and when to consider a PyPy version really outdated? But this PR seems to just be offering them all, as it serves a slightly different purpose to `cibuildwheel`.
> https://downloads.python.org/pypy/versions.json shows less versions than https://downloads.python.org/pypy/, but maybe that's on purpose?

I think `https://downloads.python.org/pypy/versions.json` should include all future versions of Python including pre-release. Otherwise, some customers who freeze exact version of PyPy could be broken when old versions are removed from `versions.json`

mattip:
> https://downloads.python.org/pypy/versions.json shows less versions than https://downloads.python.org/pypy/, but maybe that's on purpose?

Yes, the starting point is different. There is little point in offering old versions of PyPy. But, as @maxim-lobanov points out, removing versions is not really an option. So with time the cibuildwheel offerings and the `versions.json` offerings should asymptotically be equivalent.

konradpabjan (approved):
Looks great and fantastic test coverage 🚀 Ready to ship IMO

maxim-lobanov (approved):

vsafonkin (approved):

AlenaSviridenko (approved):

maxim-lobanov:
Changes from this PR were released as `actions/setup-python@v2`.
Please pay attention that currently, new syntax will install ignore image cache and always install on-flight due to lack of PYPY_VERSION files on Hosted images.
We have prepared PR to add them to Hosted images https://github.com/actions/virtual-environments/pull/2285. With next image deployment, PyPy will be started to use from image cache

nulano:
The following step:
```yml
    - name: Set up Python
      uses: actions/setup-python@v2
      with:
        python-version: pypy-3.6
        architecture: x64
```
outputs
```
PyPy version 3.6 (x) was not found in the local cache
Downloading PyPy from "https://downloads.python.org/pypy/pypy3.6-v7.3.3-win32.zip" ...
[...]
```
Is this intentional? I was expecting it to fail (the run above was caused by a typo in the matrix) since there is no 64-bit version of PyPy for Windows yet.

maxim-lobanov:
@nulano , it is expected due to this condition: https://github.com/actions/setup-python/blob/main/src/find-pypy.ts#L31
This condition is needed because all GitHub Hosted machines are x64 and when you don't specify arch explicitly, the arch of OS is used. So we fallback to x86 for Windows

nulano:
> @nulano , it is expected due to this condition: https://github.com/actions/setup-python/blob/main/src/find-pypy.ts#L31
> This condition is needed because all GitHub Hosted machines are x64 and when you don't specify arch explicitly, the arch of OS is used. So we fallback to x86 for Windows

@maxim-lobanov This is now breaking installation for PyPy 7.3.4 rc1 and rc2: #196.

PR #169, opened 2020-12-16
Improve find-python to add "Scripts" folder to PATH on Windows machines

nikita-bykov:
In scope of this PR we added `Scripts` folder to PATH on Windows. Because starting from PyPy 7.3.1 this folder using for pip and anything that pip installs on Windows. Now we have to [create symlink](https://github.com/actions/virtual-environments/blob/main/images/win/scripts/Installers/Install-PyPy.ps1#L72) on image side to avoid issues and we want to get rid of it by this changes.

### DIFF
--- a/src/find-python.ts
+++ b/src/find-python.ts
@@ -66,6 +66,10 @@ function usePyPy(
 
   core.addPath(installDir);
   core.addPath(_binDir);
+  // Starting from PyPy 7.3.1, the folder that is used for pip and anything that pip installs should be "Scripts" on Windows.
+  if (IS_WINDOWS) {
+    core.addPath(path.join(installDir, 'Scripts'));
+  }
 
   const impl = 'pypy' + majorVersion.toString();
   core.setOutput('python-version', impl);
--- a/src/find-python.ts
+++ b/src/find-python.ts
@@ -71,6 +71,11 @@ function usePyPy(
     core.addPath(path.join(installDir, 'Scripts'));
   }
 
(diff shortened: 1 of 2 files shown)


konradpabjan (approved):

src/find-python.ts
```diff
@@ -66,6 +66,15 @@ function usePyPy(
...
   core.addPath(_binDir);
+  // Starting from PyPy 7.3.1, the folder that is used for pip and anything that pip installs should be "Scripts" on Windows.
+  if (IS_WINDOWS) {
+    core.addPath(path.join(installDir, 'Scripts'));
+  }
+
+  if (IS_WINDOWS) {
+    const scriptDir = path.join(installDir, 'Scripts');
+    core.addPath(scriptDir);
+  }
```
konradpabjan, line 77: Both blocks appear to be doing the same thing? This can just be simplified to
```
  // Starting from PyPy 7.3.1, the folder that is used for pip and anything that pip installs should be "Scripts" on Windows.
  if (IS_WINDOWS) {
    const scriptDir = path.join(installDir, 'Scripts');
    core.addPath(scriptDir);
  }
```

konradpabjan (changes requested):

src/find-python.ts
dmitry-shibanov: Thank you. Removed duplicate block.

konradpabjan (approved):

Issue #196, opened 2021-03-31
PyPy 7.3.4 fails to install on Windows


nulano:
**Describe the bug**
PyPy 7.3.4 doesn't work on Windows, likely because the action is trying to install the 32-bit version (which does not exist) even when 64-bit is specified (which does exist).

**Which version of the action are you using?**

- [ ] `v1`
- [X] `v2`
- [ ] Some other tag (such as `v2.0.1` or `master`)

**Environment**
- [ ] self-hosted
- [ ] Linux
- [X] Windows
- [ ] Mac

**Python Versions**
`pypy-3.7-v7.3.4`, `pypy-3.7-v7.3.4rc1`, `pypy-3.7-v7.3.4rc2`

**To Reproduce**
```yml
    strategy:
      matrix:
        architecture: ["x86", "x64"]
    steps:
    - uses: actions/setup-python@v2
      with:
        python-version: pypy-3.7-v7.3.4rc1
        architecture: ${{ matrix.architecture }}
```

Gives the following error for both architectures:
```
Run actions/setup-python@v2
  with:
    python-version: pypy-3.7-v7.3.4rc1
    architecture: x64
    token: ***
PyPy version 3.7 (v7.3.4-rc.1) was not found in the local cache
Error: PyPy version 3.7 (v7.3.4-rc.1) with arch x86 not found
```

**Run/Repo Url**
https://github.com/nulano/Pillow/runs/2235978354?check_suite_focus=true

**Additional context**
This is likely caused by the fact this action assumes PyPy is 32-bit only on Windows, while in reality starting with version 7.3.4 PyPy is 64-bit only on Windows. 
Related to https://github.com/actions/setup-python/pull/168#issuecomment-747645831

nulano:
PyPy 7.3.4rc2 is also not found:
```
Run actions/setup-python@v2
  with:
    python-version: pypy-3.7-v7.3.4rc2
    architecture: x64
    token: ***
PyPy version 3.7 (v7.3.4-rc.2) was not found in the local cache
Error: PyPy version 3.7 (v7.3.4-rc.2) with arch x86 not found
```

nulano:
The release is also not found:
```
Run actions/setup-python@v2
  with:
    python-version: pypy-3.7-v7.3.4
    architecture: x64
    token: ***
PyPy version 3.7 (v7.3.4) was not found in the local cache
Error: PyPy version 3.7 (v7.3.4) with arch x86 not found
```

[...]

PR #201, opened 2021-04-08
Fix PyPy installation on Windows to adopt new parameters format

AlenaSviridenko:
Issue: https://github.com/actions/setup-python/issues/196

PyPy updated release parameters for Windows: `x86` -> `x64`, `win32` -> `win64` and new versions cannot be found because of it. 
I added support of 2 possible arch and platforms values. Also, added tests.

### DIFF
--- a/.github/workflows/test-pypy.yml
+++ b/.github/workflows/test-pypy.yml
@@ -18,13 +18,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-18.04, ubuntu-20.04]
+        os: [windows-latest, ubuntu-20.04]
         pypy:
         - 'pypy-2.7'
-        - 'pypy-3.6'
-        - 'pypy-3.7'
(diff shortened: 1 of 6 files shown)


.github/workflows/test-pypy.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         - 'pypy-3.7-v7.x'
         - 'pypy-3.6-v7.3.3rc1'
         - 'pypy-3.7-nightly'
+        - 'pypy-2.7-v7.3.4rc1'
```
maxim-lobanov, line 33: May be just replace `pypy-3.6-v7.3.3rc1` to avoid growing test matrix

maxim-lobanov (approved):

dmitry-shibanov (approved):

.github/workflows/test-pypy.yml
AlenaSviridenko: makes sense, updated matrix

konradpabjan (approved):

PR #263, opened 2021-11-05
Update README.md

earonesty:
dashes are not valid in matrix variables

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,14 +41,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


hugovk:
They work for me: https://github.com/python-pillow/Pillow/blob/24f0bbf5ecd2b4bc230de7ad8f5c23c7c8377fdc/.github/workflows/test.yml#L15-L44

Please can you link to your non-working config file?

nikita-bykov:
Hello @earonesty!
Dashes in matrix variables works for me as expected. Could you please provide repro steps or a test repository where the problem is observed?

Issue #229, opened 2021-08-06
Not using "token" on github server


ecs-jnguyen:
**Describe the bug**

Hi I am using actions/setup-python on a private github server. When I run `curl https://api.github.com/rate_limit` in the self-hosted runner I have 60 remaining resources. When I run the curl command again after `uses: actions/setup-python` I see that the resource counter goes down to 58. Since I am running this action a lot I eventually exceed my api call limit.

I tried using the key `token` under the with section with a token I created from github cloud but that did not work. I saw that when we are running on github server we are setting the `AUTH` variable to undefined. [Link to code](https://github.com/actions/setup-python/blob/main/src/install-python.ts#L9).

I know we are using `${{ github.token }}` by default from the `actions.yml` file. So I believe the check for github server was added so that tokens from the private github server don't get sent to github cloud. Could we add a second input for the token so that people running on private github servers don't run into the api call limit error? Or if we detect that we are running on a private github server, could we try to use `actions/python-versions` on the github server instead?

**Which version of the action are you using?**

- [ ] `v1`
- [x] `v2`
- [ ] Some other tag (such as `v2.0.1` or `master`)

**Environment**
- [ x ] self-hosted
- [ ] Linux
- [ ] Windows
- [ ] Mac

If applicable, please specify if you're using a container

**Python Versions**
 - 3.8
 - 3.9

**To Reproduce**
1. Run github actions with `actions/setup-python@v2` on github server a lot
2. Then we run into `Error: API rate limit exceeded for <IP address>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)`

github actions yaml file:
```yaml
name: CI

on:
  push:
    branches: [ master ]

jobs:
  sample_build:
    runs-on: [ self-hosted ]
    steps:
      - name: "rate limit"
        run: |
          curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/rate_limit

      - name: Set up Python 3.9
        uses: actions/setup-python@v2
        with:
          token: ${{ secrets.GH_CLOUD_TOKEN }}
          python-version: 3.9

      - name: "rate limit"
        run: |
          curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/rate_limit

```

brcrista:
When people are hitting the rate limit, is it from a single self-hosted runner being reused for multiple jobs that call `actions/setup-python`?  Or is it possible that multiple self-hosted runners are interacting with each other?

[...]

PR #316, opened 2022-01-10
Add support for adding a token on GHES to prevent rate limiting

HenrikPoulsen:
**Description:**
Adding a new input to specify a token when running on GHES.
When running on GHES today `token` gets thrown away even when explicitly set to be a github.com token.
So there is no good way to avoid the rate limiting that would occur.

No idea if this change is idiomatic or not since I don't usually work with typescript. 
So let me know if there's any changes you would want me to make.

**Related issue:**
#229 

**Check list:**

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -16,6 +16,9 @@ inputs:
     default: ${{ github.token }}
   cache-dependency-path:
     description: 'Used to specify the path to dependency files. Supports wildcards or a list of file names for caching multiple dependencies.'
+  ghes_token:
+    description: Used to pull python distributions from actions/python-versions when using Github Enterprise. This should be a github.com read only access token
+    default: ""
 outputs:
   python-version:
     description: "The installed python version. Useful when given a version range as input."
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -52264,7 +52264,12 @@ const tc = __importStar(__webpack_require__(533));
 const exec = __importStar(__webpack_require__(986));
 const utils_1 = __webpack_require__(163);
 const TOKEN = core.getInput('token');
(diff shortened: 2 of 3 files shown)


brcrista (changes requested):
We've discussed different approaches to this problem.  This approach should work, but requiring users to create a PAT in github.com to then use in GHES would create more problems than it would solve.

I think the best approach to unblock you is to fork the action on your GHES instance and apply these changes.

HenrikPoulsen:
What about fixing this for everyone that isn't me then?
Will something be done to fix this rate limiting issue? This PR at least proposes a generic workaround. Is it a hassle to set up? Sure
Does it solve a problem for people who are straight up blocked? Yes
Just making sure this isn't the end of the road for fixing this issue permanently

lasse-aagren:
The Solution implemented in this PR will continue to work exactly as today, for users not setting up til github.com PAT in their GHE project.

It will work until they hit a rate limit.

This PR provides a solution to that, for those interested. Isn't that to prefer compared to not have a solution?

brcrista:
It doesn't seem like a common issue so far, but we have thought about it.

Having to add a GHES-specific input to an action suggests to me that there's a smoother flow we should look for on the backend.  The solution we would like to implement is to make this flow part of [GitHub Connect](https://docs.github.com/en/enterprise-server@3.3/admin/configuration/configuring-github-connect/managing-github-connect).  That would solve the problem in a general way for all first-party actions and be consistent with other flows that fit this pattern.  No ETA on this, though.

Vercasek1985:
Hello @brcrista
I am working in the automation team. We do the automation in ansible. Github is used heavily in our company (Kyndryl) and we really need this fix ASAP. Could you please help me, what can we do to prioritize this issue to be fixed in very near future?

Issue #38, opened 2019-11-10
Release setup-python as an npm package


peter-evans:
I have a Javascript action that executes Python. It would be great if I could add `@actions/setup-python` as a dependency and call it from my Javascript action.

In order to make it work for now I converted parts of this action into a `setupPython` function that I've included with my action. Below is a simplified snippet of what I'm doing now. [Full source code here](https://github.com/peter-evans/create-pull-request/blob/master/index.js).
```javascript
async function run() {
  try {
    const src = __dirname + "/src";

    // Setup Python from the tool cache
    setupPython("3.8.0", "x64");

    // Install requirements
    await exec.exec("pip", [
      "install",
      "--requirement",
      `${src}/requirements.txt`
    ]);

    // Execute python script
    await exec.exec("python", [`${src}/create-pull-request.py`]);
  } catch (error) {
    core.setFailed(error.message);
  }
}
```

[Here is a python-action template](https://github.com/peter-evans/python-action) that also demonstrates what I'm doing in the snippet above.

**Why call Python from a Javascript action?**
The reason I ended up doing this and not writing the whole action in Javascript requires a bit of history and explanation. I wrote the action during GitHub Actions beta v1 when it was just container actions. When Actions v2 arrived the choices were Javascript or Docker container actions. I want the action to be multi-platform and runnable on any virtual machine, so I have no choice but to wrap it with a Javascript action as I'm doing, or convert the whole action to Javascript.

Even if Docker container actions worked on all platforms I think I would still choose to wrap Python in Javascript like this. The two issues I don't like about container actions are:
- When using `image: 'Dockerfile'` it's slow because the image is built from scratch every time the workflow runs.
- When using `image: 'docker://my-namespace/my-image:1.0.0'` it cannot be forked easily because the reference to the public Docker image remains. Being able to fork GitHub actions is important for security conscious users.

kevinresol:
Or it should support triggering another action from a javascript action, e.g.

```js
import * as actions from '@actions/actions'; // <-- this is new
import * as exec from '@actions/exec';
await actions.run('setup-python', {python-version: '2.x'});
await exec.exec('pip install ...');
```

peter-evans:
It's not particularly pretty, but I think I've found a way to include `setup-python` as a dependency and execute it.

Add the repository as a dependency and specify a version.
```
npm i --save actions/setup-python#v2.0.1
```
It will get added to package.json like this.
```
  "dependencies": {
[...]

PR #329, opened 2022-02-01
Add `build-npm` script

bexnoss:
**Description:**
Adds a `build-npm` script and documents it. This basically implements the workaround from https://github.com/actions/setup-python/issues/38#issuecomment-634471245 with added declaration files as a npm script.

The `build-npm` script does the following things:
- `npm install` since in this use case the project is installed in node_modules it is necessary to install first
- `node -e "require('fs').rmSync('lib', { recursive: true, force: true })"` is `rm -rf lib` in Node. This is necessary because tsc thinks the generated declaration files in lib are source files and refuses to overwrite them.
- `tsc --declaration --declarationMap` builds the project with declarations into lib

**Related issue:**
https://github.com/actions/setup-python/issues/38

**Check list:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "main": "dist/index.js",
   "scripts": {
     "build": "ncc build -o dist/setup src/setup-python.ts && ncc build -o dist/cache-save src/cache-save.ts",
+    "build-npm": "npm ci && node -e \"try{require('fs').rmdirSync('lib',{recursive:true})}catch(e){}\" && tsc --declaration --declarationMap",
     "format": "prettier --write \"{,!(node_modules)/**/}*.ts\"",
     "format-check": "prettier --check \"{,!(node_modules)/**/}*.ts\"",
     "release": "ncc build -o dist/setup src/setup-python.ts && ncc build -o dist/cache-save src/cache-save.ts && git add -f dist/",


dmitry-shibanov:
Hello @bexnoss. Thank you for your pull request. Can we use `npm ci` instead of `npm install` command ? As I know dependencies in package-lock.json will be updated through the npm install command and they won't be the same as for setup-python in `dist`.
Could you please rewrite `node -e \"require('fs').rmSync('lib', { recursive: true, force: true })\"` ? My main concern that the support for `rmSync` function was added in node 14, that is why it will fail on older versions.

bexnoss:
Hello @dmitry-shibanov, `npm install` is required in npm v7 to install dependencies from GitHub (see https://github.com/npm/cli/issues/2610) but since setup-python itself doesn't install any dependencies from GitHub it can be changed to `npm ci`.

I'll also change `rmSync` to `rmdirSync`.

bexnoss:
I have removed the documentation from the README and instead put it into the commit message. This way determined users can still find it but it isn't advertised as an intended use case.

The goals of this PR are to add a "correct" way of building @actions/setup-python as a NPM dependency so users don't have to experiment with various hacks and also to future proof this, so if the build process gets changed again the "build-npm" script can stay unchanged and keep working for this use case.

Please let me know if there is anything else that can be improved.

dmitry-shibanov (approved):

dmitry-shibanov:
Hello @bexnoss. Sorry for the late response. Could you please sync with the main branch ?

marko-zivic-93 (approved):

dmitry-shibanov (approved):

brcrista (changes requested):
Thanks for the PR @bexnoss.  I don't think publishing as an NPM package is the direction we want to go in.  I'll comment on the issue, but I think composite actions is the right approach to reusability here.

dmitry-shibanov:
Hello everyone. For now I'm going to close the pull request because as mentioned the composite action should help.

PR #336, opened 2022-02-10
add support for python-version-file

### LINKED ISSUE #291: Support a `.python-version` file
**Description:**
The [2.5.0 release of _actions/setup-node_](https://github.com/actions/setup-node/releases/tag/v2.5.0) released support determine the node version via an `.nvmrc` file at the root. I think it'd be prudent to use the [_pyenv_ standard of a `.python-version` file](https://github.com/pyenv/pyenv#choosing-the-python-version) to support this.

**Justification:**
It'd be nice to be DRY when specifying versions across multiple actions.

adilosa:
**Description:**
Support for a `python-version-file` input to read files such as `.python-version`, similar to setup-node and setup-ruby.

Since `python-version` has a default of '3.x' it is always specified - and altering that would be a breaking change - here `python-version-file` overrides `python-version`. This is opposite to actions/setup-node. 

**Related issue:**
Fixes #291 

**Check list:**
- [x] Mark if documentation changes are required. _(README was updated)_


.github/workflows/test-python.yml
```diff
@@ -62,6 +63,39 @@ jobs:
...
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: build-version-file ${{ matrix.python }}
+      run: echo ${{ matrix.python }} > .python-version
+
+    - name: setup-python ${{ matrix.python }}
+      uses: ./
+      with:
+        python-version-file: '.python-version'
```
lukasz-mitka, line 84: I would expect to specify this external version source like so
```suggestion
    - name: setup-python (version from file)
      uses: ./
      with:
        python-version: 'file'
```


With file name being customizable and default=`.python-version`
```suggestion
    - name: setup-python ${{ matrix.python }}
      uses: ./
      with:
        python-version: 'file'
        python-version-file: '.custom-python-version'
```

action.yml
```diff
@@ -6,6 +6,8 @@ inputs:
   python-version:
     description: "Version range or exact version of a Python version to use, using SemVer's version range syntax."
     default: '3.x'
+  python-version-file:
+    description: "File containing the Python version to use.  Examples: .python-version"
```
lukasz-mitka, line 10:
```suggestion
    description: "File containing the Python version to use.  Examples: .python-version"
    default: ".python-version"
```

lukasz-mitka (commented):
I would expect different approach - specify `python-version: 'file'` to use file as version source.

adilosa:
I think the presence of the `python-version-file` is enough of a signal to use the file, without also having to set an additional magic value to `python-version`. This also keeps it more consistent with the other actions/setup-node and actions/setup-ruby actions.

lukasz-mitka:
> I think the presence of the `python-version-file` is enough of a signal to use the file, without also having to set an additional magic value to `python-version`. This also keeps it more consistent with the other actions/setup-node and actions/setup-ruby actions.

Ok, I see what you are trying to achieve.

`actions/setup-node` uses conflict resolution which is missing here https://github.com/actions/setup-node/blob/main/src/main.ts#L80
Also `python-version-file` should have description that it's inferior to `python-version`.

Ruby has a nice 'default to file' if version is unspecified which could be handy, but should be a separate PR https://github.com/ruby/setup-ruby/blob/master/action.yml#L9

.github/workflows/test-python.yml
lukasz-mitka: Resolved

action.yml
lukasz-mitka: Resolved.

adilosa:
> actions/setup-node uses conflict resolution which is missing here https://github.com/actions/setup-node/blob/main/src/main.ts#L80

Right - I tried that at first, and would have kept it since I was mostly copying that code anyway. But we can't do that here because `python-version` has a default of '3.x' specified. This means the variable is literally always set and cannot be unset, so a conflict would always happen when `python-version-file` is specified. Not a huge fan of this, but as stated in the PR comment altering it would be a breaking change.

>  Also python-version-file should have description that it's inferior to python-version.

Yep, agreed. I'll update that.

> Ruby has a nice 'default to file' if version is unspecified which could be handy, but should be a separate PR https://github.com/ruby/setup-ruby/blob/master/action.yml#L9

Also agreed. If it weren't for `python-version` having a default specified, this would make more sense IMO and be more consistent. Unfortunately will have to be another PR and probably major version increase I suspect.

dsame:
@adilosa Hi Andrew, can you please clarify the current status of the PR? Do you want me to help you with it? Namely i can resolve the conflicts with the fork of your branch - should i?

adilosa:
Hi, I think it's just waiting on review. It seems to have gotten out of date in the meantime.

adilosa:
I've updated it with the latest upstream changes and rebuilt so there's no longer a merge conflict. Lmk if there's anything else.

marko-zivic-93 (approved):

src/setup-python.ts
```diff
@@ -21,6 +22,27 @@ async function cacheDependencies(cache: string, pythonVersion: string) {
   await cacheDistributor.restoreCache();
 }
 
+function resolveVersionInput(): string {
```
dmitry-shibanov, line 25: I think python-version input should take precedence over the python-version-file. It should be similar to this one https://github.com/actions/setup-node/blob/main/src/main.ts#L66-L97

dmitry-shibanov:
Hello @adilosa. Could you please add unit tests for the `python-version-file` input.

src/setup-python.ts
adilosa: Right, yes, but as noted this would be a breaking change due to the YAML default of `python-version: '3.x'`. This is a difference from setup-node. It means `python-version` is _always_ specified and it cannot be unspecified. So, by the setup-node logic, `python-version-file` would never be used no matter what the user does. 

If we wanted to remove the problematic default and require the user to specify at least one of `version` or `version-file`, like setup-node does, then the action would start crashing for existing setups who relied on that default. I would guess that's most of them.

I agree that setup-node makes more sense, and parity between the setup-* actions is desirable. I think to do that we'd have to bump major version and document the difference and config migration step. I'm not sure what the procedure is for such changes here, is that what you'd go with?

.github/workflows/test-python.yml
```diff
@@ -62,6 +63,39 @@ jobs:
...
+    name: Setup ${{ matrix.python }} ${{ matrix.os }} version file
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-18.04, ubuntu-20.04]
+        python: [3.5.4, 3.6.7, 3.7.5, 3.8.1]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
```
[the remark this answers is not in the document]
adilosa: Sure. This comes from the other tests in this file which all use v2, should I still update this one? or all of them?

src/setup-python.ts
```diff
@@ -21,6 +22,27 @@ async function cacheDependencies(cache: string, pythonVersion: string) {
...
+      throw new Error(
+        `The specified node version file at: ${versionFilePath} does not exist`
+      );
+    }
+    version = fs.readFileSync(versionFilePath, 'utf8');
+    core.info(`Resolved ${versionFileInput} as ${version}`);
+  }
+
+  return version;
+}
```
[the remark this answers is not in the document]
adilosa: Dimitry mentioned the same, I replied in that thread :) 

https://github.com/actions/setup-python/pull/336#discussion_r854545710

src/setup-python.ts
dmitry-shibanov: You're right. The action has default value for the python-version input, that is why it won't be empty string. I think we can rely on python-version-file with precedence.

dmitry-shibanov (approved):

src/setup-python.ts
lukasz-mitka: @dmitry-shibanov @adilosa  I would say it might be better to introduce breaking change (and do a major release) to align with  `setup-node` and `setup-ruby` actions.
With current solution `setup-python` would end up in minority and potentially confuse users.

.github/workflows/test-python.yml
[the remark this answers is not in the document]
vsafonkin: All of them, I believe

src/setup-python.ts
dsame: > It means `python-version` is _always_ specified

@adilosa i have a rather cumbersome idea how to make the python-version high priority without breaking the existing builds

1. replace "default value" and for input with some string like "default" or "not-set"
2. Handle that string by code, resolving to the the "3.x" python version

As a result:

- if you have python-version set to some meaningful value - use it as version despite of python-version-file
- if you have python-version set to the "default" value and python-version-file is not set - use  "3.x"
- if you have python-version set to the "default" value and python-version-file is set - use python-version-file

Did not i miss something?

brcrista: Nice idea @dsame.  I was thinking that perhaps the input value is retrievable from the `steps` context, but it looks like that's only the output: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context

It's still a papercut that `setup-python` has a default value while other `setup-` actions don't, but I think that's a solution to get similar behavior without making a breaking change.

brcrista: If it comes down to it, I'd be in favor of just doing a major version update and getting `setup-python` in sync with the other `setup-` actions.

dsame: > If it comes down to it, I'd be in favor of just doing a major version update and getting `setup-python` in sync with the other `setup-` actions.

Hello @brcrista , so as we all seem to agree to make breaking change and bump major version, please confirm your approve, and either @adilosa  or me will complete this piece of work.

brcrista: Thanks @dsame, sounds good to me.

brcrista (approved):

vsafonkin (approved):

marko-zivic-93 (approved):

src/setup-python.ts
dsame: @adilosa will you make the breaking change?

adilosa: sounds good - I've updated the PR with the new changes. it should be like the other `setup-` actions now, which is nice.

adilosa:
Yep, no problem. I've reverted the README change

vsafonkin (approved):

marko-zivic-93 (approved):

adilosa:
resolved the merge conflict. LGTM!

vsafonkin:
Hi @adilosa, could you please update your branch from `main` to fix poetry cache test?

adilosa:
@vsafonkin yep, should be updated now

vsafonkin:
@adilosa, could you please rebuild javascript files?

adilosa:
updated. I had rerun the build already, but apparently I needed to update the installed dependencies as well.

Issue #162, opened 2020-11-13
Error: Version X with arch x64 not found


kldavis4:
**Describe the bug**
I am attempting to use setup-python with a self-hosted runner running on an ec2 instance. I am using this ami `ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-20201112`. I have followed the instructions [here](https://github.com/actions/setup-python#linux). I am getting the following output when the setup-python action is run:

```
Run actions/setup-python@v2
  with:
    python-version: 3.7.7
    token: ***
  env:
    AGENT_TOOLSDIRECTORY: /opt/hostedtoolcache
Version 3.7.7 was not found in the local cache
Version 3.7.7 is available for downloading
Download from "https://github.com/actions/python-versions/releases/download/3.7.7-20200609.4/python-3.7.7-linux-18.04-x64.tar.gz"
Extract downloaded archive
/bin/tar xz --warning=no-unknown-keyword -C /actions-runner/_work/_temp/f7f341f3-b2b9-4eff-bff0-5b55acb3a0f1 -f /actions-runner/_work/_temp/402c2129-71c1-46e4-9263-3b04a4b9772a
Execute installation script
Check if Python hostedtoolcache folder exist...
Create Python 3.7.7 folder
Copy Python binaries to hostedtoolcache folder
Create additional symlinks (Required for the UsePythonVersion Azure Pipelines task and the setup-python GitHub Action)
Upgrading PIP...
Looking in links: /tmp/tmpryj40n3d
Requirement already satisfied: setuptools in /opt/hostedtoolcache/Python/3.7.7/x64/lib/python3.7/site-packages (41.2.0)
Requirement already satisfied: pip in /opt/hostedtoolcache/Python/3.7.7/x64/lib/python3.7/site-packages (19.2.3)
Collecting pip
Using cached https://files.pythonhosted.org/packages/cb/28/91f26bd088ce8e22169032100d4260614fc3da435025ff389ef1d396a433/pip-20.2.4-py2.py3-none-any.whl
Installing collected packages: pip
Successfully installed pip-20.2.4
Create complete file
Error: Version 3.7.7 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

It's unclear what is actually failing as it seems to be finding the 3.7.7 python version and installs it and then it fails after installing pip.

**Which version of the action are you using?**

- [ ] `v1`
- [X] `v2`
[...]

PR #90, opened 2020-05-08
Updated self-hosted guidelines for Linux and Mac

konradpabjan:
Automatic self-hosted setup with `v2` will not work out of the box on Mac and Linux unless users perform some extra steps on their machines.

This updates the `README` so users know what they have to do.

🎨 [Rendered](https://github.com/actions/setup-python/blob/konradpabjan/self-hosted-linux-mac/README.md)🖌

PR #338, opened 2022-02-17
Force TOOLCACHE_ROOT to be equal AGENT_TOOLSDIRECTORY

dsame:
**Description:**
The AGENT_TOOLSDIRECTORY variable tells where python has to be installed to
The RUNNER_TOOL_CACHE variable tells where python expected to be installed in

If these 2 variables are different the `actions/tool-cache.find` does not find a python instance installed by `actions/setup-pyton.installCpythonFromRelease`

Test build for self-hosted agent with AGENT_TOOLSDIRECTORY set(the original issue condition): https://github.com/akv-demo/setup-python-test/runs/5237321231?check_suite_focus=true

Test for builds for hosted agents (with AGENT_TOOLSDIRECTORY set and omitted)

https://github.com/akv-demo/setup-python-test/runs/5237321104?check_suite_focus=true
https://github.com/akv-demo/setup-python-test/runs/5237321426?check_suite_focus=true

Upd: last changes sample builds https://github.com/akv-demo/setup-python-test/actions/runs/1874490317

**Note:**
It seems we should not apply this fix but remove the mention of AGENT_TOOLSDIRECTORY from the documentation. In this case the python will be installed into `_work/_tool/Python/3.7.7/x64` directory

see https://github.com/akv-demo/setup-python-test/runs/5237321176?check_suite_focus=true

**Related issue:**
[Add link to the related issue.](https://github.com/actions/setup-python/issues/162)

**Check list:**


nialloh23:
Just wondering if there is an idea of when this PR will be merged? And if it fixes the issues seen setting up pyhon versions on self hosted runners? Thanks a mill

Nick-Yazdani:
> Just wondering if there is an idea of when this PR will be merged? And if it fixes the issues seen setting up pyhon versions on self hosted runners? Thanks a mill

Yeah having this same issue and it would be great to get this PR merged

dmitry-shibanov (approved):

brcrista (changes requested):
Ooooh ... this is confusing.  Here's how these variables work:
- The runner looks for a [number of variables](https://github.com/actions/runner/blob/28f002793854c69cc666029e81061f54002ae927/src/Runner.Common/HostContext.cs#L262) for historical reasons, but prefers `RUNNER_TOOL_CACHE`.
- The script that pre-populates the tools cache with Python versions [works with both](https://github.com/actions/python-versions/blob/7e0253ff52984c54607f0f79a1fbbf0f9b926fde/installers/nix-setup-template.sh#L13).
- The actions toolkit [looks for `RUNNER_TOOL_CACHE`](https://github.com/actions/toolkit/blob/a1b068ec31a042ff1e10a522d8fdf0b8869d53ca/packages/tool-cache/src/tool-cache.ts), [not `AGENT_TOOLSDIRECTORY`](https://github.com/actions/toolkit/search?q=AGENT_TOOLSDIRECTORY).  This is what `setup-python` uses to get the version.

> It seems we should not apply this fix but remove the mention of AGENT_TOOLSDIRECTORY from the documentation. In this case the python will be installed into `_work/_tool/Python/3.7.7/x64` directory

I think you're right about this.  Some relevant discussion from when the docs were written: https://github.com/actions/setup-python/pull/90#discussion_r422323846

More specifically, **I think the right fix is to replace `AGENT_TOOLSDIRECTORY` with `RUNNER_TOOL_CACHE` in the README**.  Could you try testing that and see if it works?

cc @konradpabjan since you might have some additional context

dsame:
@brcrista 
Hello Brian, it seems i got the idea why the variable  AGENT_TOOLSDIRECTORY  was introduced and we need to merge this PR.

This variable is useful if an user wants to install the python to the other location than the default `_work/_tool/Python/3.7.7/x64` 

I [tried](https://github.com/akv-demo/setup-python-test/actions/runs/2108628255) to set `RUNNER_TOOL_CACHE` instead of `AGENT_TOOLSDIRECTORY` as you advised but it is ignored and the python is installed in its default location `_work/_tool/Python/3.7.7/x64`

Summary: i am 90% sure we need both apply the fix of this PR and make a change in the docs to make clear AGENT_TOOLSDIRECTORY is an optional variable and should be used only if an user wants to change the default locations

brcrista (approved):
Hm, ok, I guess there's something I'm missing then.  Let's go ahead and merge this.

Using the legacy variable isn't ideal, but the most important thing is to have it documented correctly.  So if you could just follow up with a change to the docs, that would be great.

dsame:
> Hm, ok, I guess there's something I'm missing then. Let's go ahead and merge this.
> 
> Using the legacy variable isn't ideal, but the most important thing is to have it documented correctly. So if you could just follow up with a change to the docs, that would be great.

And the last note - i've checked the existing README and the wording "The Python packages that are downloaded from `actions/python-versions` are originally compiled from source in `/opt/hostedtoolcache/` ... which makes them non-relocatable." explains the AGENT_TOOLSDIRECTORY is not optional and nothing in README should be changed. So we end up with merging this PR.

brcrista:
In that case, I wonder if it would just make sense to force `AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache/` in this action?

Anyway @dsame @dmitry-shibanov feel free to merge when ready.

marko-zivic-93 (approved):

dsame:
> In that case, I wonder if it would just make sense to force `AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache/` in this action?
> 
> Anyway @dsame @dmitry-shibanov feel free to merge this PR when you're ready.

helllo @brcrista ,

I agree  the current configuration is overcomplicated, but i asked to @marko-zivic-93 to merge this PR asap because we have many related issue.

I'll create the another PR soon to eliminate we need of the set the AGENT_TOOLSDIRECTORY for self-hosted agents

Issue #362, opened 2022-03-28
Caching Support for GHES


tiwarishub:
**Description:**
👋  Team,  we have recently added support for caching in GHE (target version 3.5). And we saw ``setup-python`` action is also using the Action cache service to cache dependency, based on user input, using the ``actions/cache`` package.  
As we have added support for caching in GHES 3.5, you can modify [this](https://github.com/actions/setup-python/blob/main/src/setup-python.ts#L14-L16) condition to use the new ``isFeatureAvailable`` [function](https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L52)  present in ``@actions/toolkit 2.0.0``  to check the presence of Actions cache service. 
You can also refer to [this](https://github.com/actions/cache/blob/c31528629d55886d56018ef0e8d0da5986e4fe31/src/utils/actionUtils.ts#L79-L93) change which we are doing for [actions/cache ](https://github.com/actions/cache/) action.

tiwarishub:
@actions/actions-service for visibility

tiwarishub:
PR is raised for this https://github.com/actions/setup-python/pull/363

marko-zivic-93:
We added caching support for GHES.
Closing the issue thread.

PR #363, opened 2022-03-29
Cache on ghes

tiwarishub:
**Description:**
This PR adds caching support in setup-python for GHES with version 3.5. It checks the presence of Actions cache service(which is used for caching dependencies in GHES and dotcom) using recent ``@actions/cache`` tookit package function i.e ``isFeatureAvailable``. This same function can be applied to dotcom scenario (github.com) so this function can be safely applied to dotcom scenario as well.
I have tested the below scenarios
1. When AC (Action cache) service is not present in GHES 
image

2. AC present in GHES. Cache save
image

3. AC present in GHES. restored saved cache.
image

4. Without choosing the cache option in GHES

image

5. On Dotcom, with the cache option chosen
save https://github.com/tiwarishub/python-caching/actions/runs/2060198424
image

restore https://github.com/tiwarishub/python-caching/runs/5741964174?check_suite_focus=true
image

6. On Dotcom, without cache option is chosen https://github.com/tiwarishub/python-caching/runs/5741986713?check_suite_focus=true
image

**Note**
v3 version of setup-python is not published to GHES 3.4 so with this we will also release v3 version of setup-python. Please let us know if there is any issue with that.

**Related issue:**
https://github.com/actions/setup-python/issues/362

**Check list:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "setup-python",
-  "version": "2.2.2",
+  "version": "2.2.3",
   "private": true,
   "description": "Setup python action",
   "main": "dist/index.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "setup-python",
-  "version": "2.2.3",
+  "version": "3.0.1",
   "private": true,
   "description": "Setup python action",
   "main": "dist/index.js",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "author": "GitHub",
   "license": "MIT",
   "dependencies": {
-    "@actions/cache": "^1.0.8",
+    "@actions/cache": "^2.0.0",
     "@actions/core": "^1.2.3",
     "@actions/exec": "^1.1.0",
     "@actions/glob": "^0.2.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "setup-python",
-  "version": "3.0.1",
+  "version": "3.1.0",
   "private": true,
   "description": "Setup python action",
   "main": "dist/index.js",
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -110,7 +110,7 @@ export function isCacheFeatureAvailable(): boolean {
       );
     } else {
       core.warning(
-        'An internal error has occurred in cache backend. Please check https://www.githubstatus.com/ for any ongoing issue in actions.'
+        'The runner was not able to contact the cache service. Caching will be skipped'
       );
     }
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,3 +1,4 @@
+import * as cache from '@actions/cache';
 import fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
@@ -98,4 +99,20 @@ export function isGhes(): boolean {
     process.env['GITHUB_SERVER_URL'] || 'https://github.com'
   );
   return ghUrl.hostname.toUpperCase() !== 'GITHUB.COM';
+}
+
+export function isCacheFeatureAvailable(): boolean {
+  if (!cache.isFeatureAvailable()) {
+    if (isGhes()) {
+      throw new Error(
+        'Caching is only supported on GHES version >= 3.5. If you are on version >=3.5 Please check with GHES admin if Actions cache service is enabled or not.'
+      );
+    } else {
+      throw new Error(
+        'An internal error has occurred in cache backend. Please check https://www.githubstatus.com/ for any ongoing issue in actions.'
+      );
+    }
+  }
+
+  return true;
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,4 +1,5 @@
 import * as cache from '@actions/cache';
+import * as core from '@actions/core';
 import fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
@@ -108,10 +109,12 @@ export function isCacheFeatureAvailable(): boolean {
         'Caching is only supported on GHES version >= 3.5. If you are on version >=3.5 Please check with GHES admin if Actions cache service is enabled or not.'
       );
     } else {
-      throw new Error(
+      core.warning(
         'An internal error has occurred in cache backend. Please check https://www.githubstatus.com/ for any ongoing issue in actions.'
       );
     }
+
+    return false;
   }
 
   return true;
--- a/.licenses/npm/@actions/cache.dep.yml
+++ b/.licenses/npm/@actions/cache.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: "@actions/cache"
-version: 1.0.8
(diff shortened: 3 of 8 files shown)


tiwarishub:
Few CI checks are failing but I believe this is failing on master so I think it ok to ignore it?

src/utils.ts
```diff
@@ -99,3 +100,19 @@ export function isGhes(): boolean {
   );
   return ghUrl.hostname.toUpperCase() !== 'GITHUB.COM';
 }
+
+export function isCacheFeatureAvailable(): boolean {
+  if (!cache.isFeatureAvailable()) {
+    if (isGhes()) {
+      throw new Error(
+        'Caching is only supported on GHES version >= 3.5. If you are on version >=3.5 Please check with GHES admin if Actions cache service is enabled or not.'
+      );
```
brcrista, line 109: Since caching is just an optimization, I feel like a warning would be more appropriate here.

brcrista (changes requested):
I think we should consider using warnings instead of errors for caching failures.

src/utils.ts
tiwarishub: we were already throwing an error in the case of GHES https://github.com/actions/setup-python/blob/main/src/setup-python.ts#L15. So i didn't change the end-user expectations. Should we change this behaviour?

bishal-pdMSFT: Agree, it will be a breaking behavior. In `actions/cache` action we do log a warning only. But here it was treated as an error earlier, may be because caching is additionally asked by user (by setting `cache` input) during nodejs setup.

brcrista: Ok, I can see this as an error since it's in line with the previous behavior.  In that case, just a small edit to the text:

```suggestion
      throw new Error(
        'Caching is only supported on GHES version >= 3.5.  If you are on a version >= 3.5, please check with your GHES admin if the Actions cache service is enabled or not.'
      );
```

Although, I'm confused here: can the Actions cache service be enabled/disabled independently?

src/utils.ts
```diff
@@ -99,3 +100,19 @@ export function isGhes(): boolean {
...
+    } else {
+      throw new Error(
+        'An internal error has occurred in cache backend. Please check https://www.githubstatus.com/ for any ongoing issue in actions.'
+      );
```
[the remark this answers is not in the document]
brcrista: Despite the conversation on the case above, I think this still stands as a place where we should warn instead of fail.

src/utils.ts
tiwarishub: yeah as the user is explicitly asking for cache, will it make more sense to throw an error in case the cache service is not there.  I can see both pros and cons for throwing errors and warnings. I am not sure of the initial thought process behind throwing an error instead of a warning(i believe mostly because a user has asked explicitly for caching).  Please let me know your thoughts? I am open to both

src/utils.ts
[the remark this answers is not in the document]
tiwarishub: Make sense. I will update the PR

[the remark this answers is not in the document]
tiwarishub: Updated the PR

bishal-pdMSFT (approved):
LGTM

dmitry-shibanov (approved):

src/utils.ts
```diff
@@ -99,3 +101,21 @@ export function isGhes(): boolean {
   );
   return ghUrl.hostname.toUpperCase() !== 'GITHUB.COM';
 }
+
+export function isCacheFeatureAvailable(): boolean {
+  if (!cache.isFeatureAvailable()) {
+    if (isGhes()) {
+      throw new Error(
+        'Caching is only supported on GHES version >= 3.5. If you are on a version >= 3.5, please check with your GHES admin if the Actions cache service is enabled or not.'
+      );
```
vsafonkin, line 110: Probably, it's not necessary to throw a error here?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "setup-python",
-  "version": "2.2.2",
+  "version": "2.2.3",
```
dmitry-shibanov, line 3: Could you please change it to the planning version of the setup-python action?

tiwarishub: @dmitry-shibanov, Sorry I am not aware of the planned version. Could please share what is planned version.

dmitry-shibanov: I think it should be 3.0.1 or 3.1.0 to be consistent with the setup-python action releases: https://github.com/actions/setup-python/releases/tag/v3.0.0

src/utils.ts
tiwarishub: As per the previous code, we were throwing an error in the GHES scenario https://github.com/actions/setup-python/blob/main/src/setup-python.ts#L15. Now we don't want to change end-user behaviour and this condition is similar to that that where workflow is running on GHES  and workflow has explicitly asked for the cache using ``cache`` option but it has not enabled the cache.  we had similar discussion here https://github.com/actions/setup-python/pull/363#discussion_r837782316

package.json
tiwarishub: True, even I found this to be wired that package.json has 2.2.2 and released tag 3.0.0 which we generally keep the same.  So I thought this is already followed to keep them not in sync. Shall I fix this?

src/utils.ts
```diff
@@ -99,3 +101,21 @@ export function isGhes(): boolean {
...
+    } else {
+      core.warning(
+        'An internal error has occurred in cache backend. Please check https://www.githubstatus.com/ for any ongoing issue in actions.'
```
brcrista, line 113: Is this message accurate?  It looks like `cache.isFeatureAvailable()` just looks for whether an environment variable is set.

https://github.com/actions/toolkit/blob/b4639928698a6bfe1c4bdae4b2bfdad1cb75016d/packages/cache/src/cache.ts#L53

brcrista: If my understanding is correct, I would say something like

```suggestion
        'This runner is not configured to use the cache service.  Caching will be skipped.'
```

tiwarishub: yes, this env var is only set when cache service is present otherwise this var will not be set. This else condition is for dotcom scenario so ideally this scenario should never occur because AC is always there in dotcom. But just to cover corner scenario i have added this

brcrista: Ok I understand better now https://github.com/actions/runner/blob/408d6c579c36f0eb318acfdafdcbafc872696501/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs

How about:

```suggestion
        'The runner was not able to contact the cache service.  Caching will be skipped.'
```

tiwarishub: Okay. Let me update the PR with changes.

tiwarishub: done

brcrista (approved):

vsafonkin (approved):

marko-zivic-93 (approved):
All issues we noticed were referenced :) Everything seems ok now :)

Issue #345, opened 2022-03-01
``python-version`` seems to be missing from outputs when installing ``PyPy``


DanielNoord:
**Description:**
[Run](https://github.com/PyCQA/pylint/runs/5376859270?check_suite_focus=true)
[Workflow](https://github.com/PyCQA/pylint/blob/main/.github/workflows/ci.yaml)

In the above job the `Restore Python Virtual Environment` step uses the following key:
```yaml
 ${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{
            needs.prepare-tests-pypy.outputs.python-key }}
```
See L479 of the linked workflow file.

However, if you look at the `key` shown on L15 of that step you see that the `python-version` is missing.

This step does work for the `cpython` jobs on both Linux and Windows, but not for `PyPy`.
Is this a known issue?

Edit: I checked and this behaviour did not occur a couple of weeks ago. Might this be related to #342?

**Action version:**
3.0.0

**Platform:**
- [x] Ubuntu

**Runner type:**
- [x] Hosted

**Tools version:**
`pypy-3.6`, `pypy-3.7` and possibly more.

**Repro steps:**  
See the run and workflow linked above.

**Expected behavior:**
The `setup-python` action to have `python-version` in its outputs when `PyPy` is installed.

**Actual behavior:**
`python-version` seems to be missing.

marko-zivic-93:
Hello @DanielNoord 
Thank you for your report!
We will take a look at this issue and come back with results as soon as possible.

dsame:
@marko-zivic-93 please review my PR

DanielNoord:
I think this can be closed now that #365 has been merged. @brcrista Gentle ping since you merged that PR 😄

PR #365, opened 2022-03-31
Add pyton-version to setup PyPy output

dsame:
**Description:**
During the migration from the [legacy PyPy input ](https://github.com/actions/setup-python/pull/342) the setting output "pyton-version" was omitted

**Related issue:**
[link to the related issue](https://github.com/actions/setup-python/issues/345)

Test build
https://github.com/akv-demo/setup-python-test/runs/5765771141?check_suite_focus=true

confirmation the output is set:
https://github.com/akv-demo/setup-python-test/runs/5765771141?check_suite_focus=true#step:4:39

**Check list:**


dmitry-shibanov:
I think it'll be better to move such kind of logic [here](https://github.com/actions/setup-python/blob/main/src/find-pypy.ts#L52-L55).

vsafonkin (approved):

dsame:
> I think it'll be better to move such kind of logic [here](https://github.com/actions/setup-python/blob/main/src/find-pypy.ts#L52-L55).

Yes, it is done,
test build: https://github.com/akv-demo/setup-python-test/runs/5801619011?check_suite_focus=true

src/find-pypy.ts
```diff
@@ -52,6 +52,7 @@ export async function findPyPyVersion(
   core.exportVariable('pythonLocation', pythonLocation);
   core.addPath(pythonLocation);
   core.addPath(_binDir);
+  core.setOutput('python-version', 'pypy' + resolvedPyPyVersion.trim());
```
brcrista, line 55: Is there a test we can add to make sure it doesn't regress again?

vsafonkin (approved):

PR #373, opened 2022-04-03
Cache hit output

### LINKED ISSUE #371: Introduce cache-hit output
**Description:**
I'd like to propose adding a new output called `cache-hit` (like in actions/cache) to show whether dependency cache was hit.

**Justification:**
Since this action utilizes caching mechanic I think it would be needed to have this meta data in output since we can't access internal steps from outside. This would be helpful if, for example, we need add stuff to `PATH` if this cache was hit.

Yes, for sure.

dhvcc:
**Description:**
- Added `cache-hit` output
- Removed useless `await`s in `cache-restore` tests

**Related issue:**
closes #371. Test action run is [here](https://github.com/dhvcc/debug-actions-repo/runs/5802597704?check_suite_focus=true)

**Check list:**
- [X] Mark if tests were added or updated to cover the changes.

### DIFF
--- a/src/cache-distributions/cache-distributor.ts
+++ b/src/cache-distributions/cache-distributor.ts
@@ -54,6 +54,7 @@ abstract class CacheDistributor {
     } else {
       core.info(`${this.packageManager} cache is not found`);
     }
+    core.info('cache was hit');
     core.setOutput('cache-hit', Boolean(matchedKey));
   }
 }
--- a/src/cache-distributions/cache-distributor.ts
+++ b/src/cache-distributions/cache-distributor.ts
@@ -1,6 +1,5 @@
 import * as cache from '@actions/cache';
 import * as core from '@actions/core';
-import {PromiseReturnType} from '../utils';
 
 export enum State {
   STATE_CACHE_PRIMARY_KEY = 'cache-primary-key',
@@ -42,19 +41,16 @@ abstract class CacheDistributor {
       restoreKey
     );
 
-    this.handleMatchResult(matchedKey);
+    this.handleMatchResult(matchedKey, primaryKey);
   }
 
-  public handleMatchResult(
-    matchedKey: PromiseReturnType<typeof cache.restoreCache>
-  ) {
-    if (matchedKey) {
+  public handleMatchResult(matchedKey: string | undefined, primaryKey: string) {
+    if (matchedKey == primaryKey) {
       core.saveState(State.CACHE_MATCHED_KEY, matchedKey);
       core.info(`Cache restored from key: ${matchedKey}`);
     } else {
       core.info(`${this.packageManager} cache is not found`);
     }
-    core.info('cache was hit');
     core.setOutput('cache-hit', Boolean(matchedKey));
   }
 }
--- a/src/cache-distributions/cache-distributor.ts
+++ b/src/cache-distributions/cache-distributor.ts
@@ -45,13 +45,13 @@ abstract class CacheDistributor {
   }
 
   public handleMatchResult(matchedKey: string | undefined, primaryKey: string) {
-    if (matchedKey == primaryKey) {
+    if (matchedKey) {
       core.saveState(State.CACHE_MATCHED_KEY, matchedKey);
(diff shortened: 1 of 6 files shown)


dhvcc:
I'm not actually sure if removing awaits belongs in this PR. I also noticed there were no tests for another output - `python-version`. So may be it makes sense to revisit tests and clean them up or add necessary ones?

dhvcc:
Seems like those fails are unrelated to code. I can take a look at the workflows themselves if people are busy

src/cache-distributions/cache-distributor.ts
```diff
@@ -41,12 +42,20 @@ abstract class CacheDistributor {
       restoreKey
     );
 
+    this.handleMatchResult(matchedKey);
+  }
+
+  public handleMatchResult(
+    matchedKey: PromiseReturnType<typeof cache.restoreCache>
```
dmitry-shibanov, line 49: I think the type should be `string | undefined`.

src/cache-distributions/cache-distributor.ts
```diff
@@ -41,12 +42,20 @@ abstract class CacheDistributor {
...
+  ) {
     if (matchedKey) {
       core.saveState(State.CACHE_MATCHED_KEY, matchedKey);
       core.info(`Cache restored from key: ${matchedKey}`);
     } else {
       core.info(`${this.packageManager} cache is not found`);
     }
+    core.info('cache was hit');
```
dmitry-shibanov, line 57: Could you please remove this core.info?

src/cache-distributions/cache-distributor.ts
```diff
@@ -41,12 +42,20 @@ abstract class CacheDistributor {
...
+    core.setOutput('cache-hit', Boolean(matchedKey));
```
dmitry-shibanov, line 58: Could you please adjust logic to check that primary key equals to matchedKey ? We use primary key for pipenv and poetry, but for pip the action has a restore key.

src/utils.ts
```diff
@@ -119,3 +119,12 @@ export function isCacheFeatureAvailable(): boolean {
 
   return true;
 }
+
+// Awaited (typescript 4.5+) polyfill. Not ideal, so use with care
+export type AwaitedPolyfill<T> = T extends PromiseLike<infer U>
```
dmitry-shibanov, line 124: I think the `cache-hit` output does not need this type.

src/cache-distributions/cache-distributor.ts
dhvcc: Oh snap, it was a bit too late hhahaha

src/utils.ts
dhvcc: That's for `matchedKey` which comes from `cache.restoreCache`. I think either this or hardcode `string?`. Whatever you prefer really

src/cache-distributions/cache-distributor.ts
dhvcc: ok cool

src/cache-distributions/cache-distributor.ts
dhvcc: Is this the check you were interested in? I've updated the PR

dmitry-shibanov: Change comparison logic for setOutput `core.setOutput('cache-hit', Boolean(matchedKey))` => `core.setOutput('cache-hit', matchedKey === primaryKey)`

dhvcc: Updated the PR. Doesn't it mean now it's a bit different logic for cache hit output and for  "cache restored from key"?

dhvcc: I get why this comparison is needed now

brcrista (approved):

dmitry-shibanov:
Hello @dhvcc. I think it should work because the action sets the keys, that is why I consider it will be enough.

dmitry-shibanov (approved):

dmitry-shibanov:
Hello @dhvcc. I think we'll release it on this week.

Issue #132, opened 2020-08-24
Inconsistent python on mac depending on bash option


moble:
**Describe the bug**
There's a really unfortunate and mostly silent bug that happens only on `macos` — not on `ubuntu` or `windows`.  I'm not sure whether this issue could be solved more easily by `setup-python` or by [`virtual-environments`](https://github.com/actions/virtual-environments), but this is where I think people will look to find this issue.  If nothing else, I feel like it should be documented here.

Somewhere, I have a fairly complicated `run` section that (so that Windows doesn't go nuts) needs bash features — but *also* the login option to bash.  Out of inertia, this became my way of running things:

```yaml
    runs-on: macos-latest
    steps:
      - uses: actions/setup-python@v2
        with:
          python-version: 3.8

      - name: Some fancy thing
        shell: bash -l {0}
        run: |
          # Some bashy fanciness
          python --version
```

But the python that gets picked up ***on macos only*** is the wrong one — i.e., not 3.8.  The reason is obvious if you're looking for it: using `bash -l` prepends `/usr/local/bin:/usr/bin` to `PATH`.  The problem is that users probably aren't looking for it; their scripts are silently using the wrong version of python.  Again, this is *not* the behavior on ubuntu or windows, both of which have their `PATH` variables ordered correctly to find the desired version of python.  It's that difference in behavior that I'm calling a bug, and I think it's clear that ubuntu and windows have  the preferable behavior.

**Which version of the action are you using?**
- [ ] `v1`
- [X] `v2`
- [ ] Some other tag (such as `v2.0.1` or `master`)

**Environment**
- [ ] self-hosted
- [ ] Linux
- [ ] Windows
- [X] Mac

If applicable, please specify if you're using a container

**Python Versions**
All versions on Mac

**To Reproduce**
Run a python command with `bash -l` as the shell on any mac instance with any python version.  A simple workflow is [here](https://github.com/moble/test_github_actions_python/blob/master/.github/workflows/test_python_versions.yml)

**Run/Repo Url**

https://github.com/moble/test_github_actions_python/actions/runs/222351275

jspaaks:
Thanks @moble, this report helped me solve my error.

maxim-lobanov:
Hello @moble , I have explained the same behavior in https://github.com/actions/setup-python/issues/179#issuecomment-750495666.
[...]

PR #405, opened 2022-05-22
feature: add a `python-path` output

### LINKED ISSUE #309: [FR] Expose an `interpreter-path` output containing the chosen Python executable path
**Description:**

$sbj

**Justification:**

In a cross-platform matrix, there's often a need to be more specific about what you invoke. Plus the command for pointing at a specific interpreter may differ. For example, I recently discovered that invoking `python3.10` which works great under *NIX fails on Windows with `D:\a\_temp\7b307349-0858-40ae-97bf-a41c70e6d07c.sh: line 1: python3.10: command not found`. And that is right after calling `setup-python` to switch to `3.10`.

It would be so much easier and precise to just have `${{ steps.use-py310.outputs.interpreter-path }}` used in place of that.

Not sure, I've got a lot on my plate now so probably not. But feel free to tag me for code review.

mayeut:
**Description:**
Expose a `python-path` output containing the chosen Python executable path.

**Related issue:**
fix #309
mitigate #132

**Check list:**

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -23,8 +23,13 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: setup default python 
+    - name: setup default python
+      id: setup-python
       uses: ./
+
+    - name: Check python-path
+      run: ./__tests__/check-python-path.sh '${{ steps.setup-python.outputs.python-path }}'
+      shell: bash
 
     - name: Validate version
       run: python --version
@@ -45,9 +50,14 @@ jobs:
       uses: actions/checkout@v2
 
     - name: setup-python ${{ matrix.python }}
+      id: setup-python
       uses: ./
       with:
         python-version: ${{ matrix.python }}
+
+    - name: Check python-path
+      run: ./__tests__/check-python-path.sh '${{ steps.setup-python.outputs.python-path }}'
+      shell: bash
 
     - name: Validate version
       run: |
@@ -74,9 +84,14 @@ jobs:
       uses: actions/checkout@v2
 
     - name: setup-python 3.9.0-beta.4
+      id: setup-python
       uses: ./
       with:
         python-version: '3.9.0-beta.4'
+
+    - name: Check python-path
+      run: ./__tests__/check-python-path.sh '${{ steps.setup-python.outputs.python-path }}'
+      shell: bash
 
     - name: Validate version
       run: |
--- a/action.yml
+++ b/action.yml
@@ -21,6 +21,8 @@ outputs:
     description: "The installed python version. Useful when given a version range as input."
   cache-hit:
     description: 'A boolean value to indicate a cache entry was found'
+  python-path:
+    description: "The absolute path to the Python executable."
 runs:
   using: 'node16'
   main: 'dist/setup/index.js'
--- a/.github/workflows/test-pypy.yml
+++ b/.github/workflows/test-pypy.yml
@@ -37,10 +37,15 @@ jobs:
         uses: actions/checkout@v2
   
       - name: setup-python ${{ matrix.pypy }}
+        id: setup-python
         uses: ./
         with:
           python-version: ${{ matrix.pypy }}
-  
+
+      - name: Check python-path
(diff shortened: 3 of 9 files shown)


action.yml
```diff
@@ -21,6 +21,8 @@ outputs:
     description: "The installed python version. Useful when given a version range as input."
   cache-hit:
     description: 'A boolean value to indicate a cache entry was found'
+  interpreter-path:
```
brcrista, line 24:
```suggestion
  python-path:
```

Naming it this way parallels the `python-version` output.

mayeut: done.

.github/workflows/test-python.yml
```diff
@@ -23,9 +23,21 @@ jobs:
...
+    - name: Check python-path
+      run: |
+        $setupPython = (${{ steps.setup-python.outputs.python-path }} -c "import sys; print(sys.executable)")
+        $pathPython = (python -c "import sys; print(sys.executable)")
+        if ("$pathPython" -ne "$setupPython"){
+          Write-Host "Executable mismatch. python in PATH is $pathPython; python-path (${{ steps.setup-python.outputs.python-path }}) is $setupPython"
+          exit 1
+        }
+        Write-Host "python-path: ${{ steps.setup-python.outputs.python-path }}"
+      shell: pwsh
```
brcrista, line 39: Why use `shell: pwsh` in this workflow and `shell: bash` in `test-pypy.yml`?  Can we consolidate on Bash?

mayeut: I started with the `test-pypy.yml` in `bash` then saw that  `shell: pwsh` was used throughout `test-python.yml` so wrote both to let the final preference/decision up to maintainers. I'll consolidate with `bash`.

mayeut:
The failure for `e2e-cache / Test poetry (Python pypy-3.7-v7.x, windows-latest)` seems unrelated to changes in this PR.

henryiii:
The output works doesn't rely on modifying the environment, and doesn't get overwritten. So you could do multiple setup-python's, and access the correct output. An example of this might be problematic:

```yaml
- uses: actions/setup-python@v3
  with:
    python-version: "3.7"
- uses: some/composite-action@v1 # Runs a different setup-python internally
# Now pythonLocation is no longer the 3.7 one!
```

.github/workflows/test-python.yml
```diff
@@ -74,10 +98,22 @@ jobs:
...
+    - name: Check python-path
+      run: |
+        PATH_PYTHON=$(python -c 'import sys; print(sys.executable)')
+        SETUP_PYTHON=$('${{ steps.setup-python.outputs.python-path }}' -c 'import sys; print(sys.executable)')
+        if [ "${SETUP_PYTHON}" != "${PATH_PYTHON}" ]; then
+          echo "Executable mismatch. python in PATH is ${PATH_PYTHON}; python-path (${{ steps.setup-python.outputs.python-path }}) is ${SETUP_PYTHON}"
+          exit 1
+        fi
+        echo 'python-path: ${{ steps.setup-python.outputs.python-path }}'
+      shell: bash
```
vsafonkin, line 115: I suggest to move this code to the `__tests__` directory as a separated bash script to avoid duplicating it many times in the test steps.

action.yml
```diff
@@ -21,6 +21,8 @@ outputs:
     description: "The installed python version. Useful when given a version range as input."
   cache-hit:
     description: 'A boolean value to indicate a cache entry was found'
+  python-path:
+    description: "The absolute path to the python interpreter executable."
```
brcrista, line 25:
```suggestion
    description: "The absolute path to the Python executable."
```

.github/workflows/test-python.yml
mayeut: done.

action.yml
mayeut: done

mayeut:
The failure for e2e-cache / Test poetry (Python pypy-3.7-v7.x, windows-latest) seems unrelated to changes in this PR.
(the previous failure on the same test went away with a relaunch of the workflow by @vsafonkin)

brcrista (approved):

vsafonkin (approved):

Issue #159, opened 2020-11-04
Modifying LD_LIBRARY_PATH causes issues


polm:
**Describe the bug**

The changes in #144 cause issues running some applications in the container, such as git cloning with https urls. See these related issues:

https://github.com/pypa/manylinux/issues/812
https://github.com/RalfG/python-wheels-manylinux-build/issues/26

**Which version of the action are you using?**

- [ ] `v1`
- [x] `v2`
- [ ] Some other tag (such as `v2.0.1` or `master`)

**Environment**
- [ ] self-hosted
- [x] Linux
- [ ] Windows
- [ ] Mac

I was using the `pypa/manylinux1_x86_64` container.

**Python Versions**

Happens with all Python versions.

**To Reproduce**

    git clone https://...

Gives a libcurl error.

**Run/Repo Url**

https://github.com/polm/fugashi/runs/1291141895?check_suite_focus=true

nikita-bykov:
Hello @polm!
This issue has been open for a year without any activity, so I'm going to close it as stale.
Please feel free to contact us if the issue is still relevant to you and we will reopen it.
Thanks!

PR #411, opened 2022-05-26
feature: add `update-environment` input

### LINKED ISSUE #410: feature: add an option not to add/update environment variables
**Description:**
Add an option not to add/update environment variables.

**Justification:**
When using `actions/setup-python`, it will update or add `PATH`, `PythonLocation`, `PKG_CONFIG` environment variables.
If one is to use `actions/setup-python` inside a composite action, those environment variables update might conflict with a user workflow.

e.g.

```yaml
- uses: actions/setup-python@v3
  with:
    python-version: "3.7"
- uses: some/composite-action@v1 # Runs a different setup-python internally with python-version: "3.10"
- run: python --version  # this will show python 3.10 while the user expects 3.7
```

With this proposal, `some/composite-action` could use:
```yaml
- uses: actions/setup-python@v3
  id: cp310
  with:
    python-version: "3.10"
    no-environment-update: true
- run: ${{ steps.cp310.outputs.python-path }} --version  # this will show python 3.10 in the composite action
```

The user workflow will now print `python 3.7` as expected.

Yes: #411

mayeut:
**Description:**
This option allows to specify if the action shall update environment variables (default) or not.
This allows to use the `setup-python` action in a composite action without side effect (except downloading/installing python if version is missing).

**Related issue:**
fix #410 

**Check list:**


mayeut:
`e2e-cache / Test poetry (Python pypy-3.7-v7.x, windows-latest)` failure has nothing to do with this PR.

brcrista (changes requested):
I'd like to explore other solutions besides adding complexity to the action.

Can you just add steps to the composite action?  For example:

```bash
# Save the environment
set > env.sh

# Restore the environment
source env.sh
```

mayeut:
@brcrista, 
there's no way to revert a `core.addPath` or a `core.exportVariable`

For exportVariable:
If they were already set, then the ones with exportVariable could be overriden back to their original values.
If they were not set, it's impossible to unset them: https://stackoverflow.com/questions/70137245/how-to-remove-an-environment-variable-on-github-actions

For addPath, it's values are always prepended to the "current" PATH, c.f. https://github.com/actions/toolkit/issues/655

vsafonkin:
Hi @mayeut, I have some concerns about this feature. In order to use python that is installed by action without $PATH updating you should use exact paths to the installed python in a composite action. If there are any other tools in composite action which use python implicitly, then these tools will not use the version you expect. This can lead to confusion about which python is used in which command or tool.

From my point of view an installation of python without adding to PATH may create more complications than solve problems.

mayeut:
> In order to use python that is installed by action without $PATH updating you should use exact paths to the installed python in a composite action. If there are any other tools in composite action which use python implicitly, then these tools will not use the version you expect. This can lead to confusion about which python version is used in some command or tool.
> From my point of view an installation of python without adding to PATH may create more complications than solve problems.

@vsafonkin,

This is an opt-in feature. It won't change a thing for existing users. If they choose to opt-in, they shall know the implications.
Furthermore, you can easily test your own action with a GHA workflow to check if it's working as expected or not but you can't possibly test how you'll break hundreds of downstream workflows using your action if you mess up the environment. I'd rather just get ahead of this most likely event by not messing with the environment in the first place.
Last time `setup-python` introduced a change in environment variables, it did create some issues (https://github.com/actions/setup-python/issues/159). I'd rather avoid that.

As said in an earlier comment, it's just impossible to restore the environment as it was before running `setup-python`. Even if it were, I feel that it would be better solved by adding a bit of complexity in this action, centralized, rather than incurring workarounds for downstream users.

vsafonkin:
@mayeut, ok, I see. Since the [AzDev `UsePython` task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/use-python-version?view=azure-devops) has a similar input `addToPath` I think it makes sense to add this feature to the action.

action.yml
```diff
@@ -17,6 +17,9 @@ inputs:
     default: ${{ github.token }}
   cache-dependency-path:
     description: 'Used to specify the path to dependency files. Supports wildcards or a list of file names for caching multiple dependencies.'
+  no-environment-update:
+    description: 'Set this option if you want the action not to update environment variables.'
+    default: false
```
brcrista, line 22:
```suggestion
  update-environment:
    description: 'Set this option if you want the action to update environment variables.'
    default: true
```

Putting boolean inputs in positive terms is usually easier for people to understand.

mayeut: done

dmitry-shibanov (approved):

marko-zivic-93 (approved):

IvanZosimov (approved):

vsafonkin (approved):

PR #414, opened 2022-06-04
feature: fallback to pre-release when no stable version is found

### LINKED ISSUE #213: How to specify best available Python minor version
I'd like to simply specify, "give me the best Python 3.10 you have". Ideally, "3.10" would indicate that specifier and include alphas and betas and release candidates until that version is stable at which point it would only include stable releases. I'd like not to have to specify "3.10.alpha - 3.10" (or similar) and later have to remove "alpha". If someone is specifying "3.10" before there's any stable release, couldn't that be simply interpreted to mean pre-releases until stable then just stable releases? It seems to me that's what 99% of users would require and would save a lot of toil around configs.

### LINKED ISSUE #501: Make 3.x match 3.x-dev in case there is no stable release yet
Current situation: We will get 3.11 in about a month and I want to test it to CI from now on.

This means I now have to use `3.11-dev` and in one month downgrade to `3.11` to get the stable release from then on.

Ideally `3.11` would match `3.11-dev` as long as there exists no stable release yet for that version, unless I'm missing something.

I know it's possible to do `3.11.0-alpha - 3.11.0` but that still would require me to clean things up eventually and I'd like to add it once and never touch it again.

mayeut:
**Description:**
This allows to specify version like `3.11` or `pypy3.10` in workflows before those versions are released.
This lessen the burden for users of `setup-python` by not having to modify their workflow twice: once when a pre-release is available (e.g. `3.11-dev`) and once when the first stable release is published (e.g. `3.11`)
Allowing `3.11` for pre-release also allows to simplify workflows in situations like in https://github.com/pypa/packaging/pull/551 which is using the `python-version` specifier as an input to nox and would require some bash string manipulation otherwise.

**Related issue:**
fixes #213
fixes #501 
https://github.com/actions/virtual-environments/issues/5540
https://github.com/actions/virtual-environments/issues/3343
https://github.com/pypa/packaging/pull/551

**Check list:**

### DIFF
--- a/__tests__/finder.test.ts
+++ b/__tests__/finder.test.ts
@@ -56,7 +56,7 @@ describe('Finder tests', () => {
     await io.mkdirP(pythonDir);
     fs.writeFileSync(`${pythonDir}.complete`, 'hello');
     // This will throw if it doesn't find it in the cache and in the manifest (because no such version exists)
-    await finder.useCpythonVersion('3.x', 'x64', true, false);
+    await finder.useCpythonVersion('3.x', 'x64', true, false, false);
     expect(spyCoreAddPath).toHaveBeenCalled();
     expect(spyCoreExportVariable).toHaveBeenCalledWith(
       'pythonLocation',
@@ -73,7 +73,7 @@ describe('Finder tests', () => {
     await io.mkdirP(pythonDir);
     fs.writeFileSync(`${pythonDir}.complete`, 'hello');
     // This will throw if it doesn't find it in the cache and in the manifest (because no such version exists)
-    await finder.useCpythonVersion('3.x', 'x64', false, false);
+    await finder.useCpythonVersion('3.x', 'x64', false, false, false);
     expect(spyCoreAddPath).not.toHaveBeenCalled();
     expect(spyCoreExportVariable).not.toHaveBeenCalled();
   });
@@ -95,7 +95,12 @@ describe('Finder tests', () => {
       fs.writeFileSync(`${pythonDir}.complete`, 'hello');
     });
     // This will throw if it doesn't find it in the cache and in the manifest (because no such version exists)
-    await finder.useCpythonVersion('1.2.3', 'x64', true, false);
+    await expect(
+      finder.useCpythonVersion('1.2.3', 'x64', true, false, false)
+    ).resolves.toEqual({
+      impl: 'CPython',
+      version: '1.2.3'
+    });
     expect(spyCoreAddPath).toHaveBeenCalled();
     expect(spyCoreExportVariable).toHaveBeenCalledWith(
       'pythonLocation',
@@ -122,14 +127,19 @@ describe('Finder tests', () => {
       const pythonDir: string = path.join(
         toolDir,
         'Python',
-        '1.2.3-beta.2',
+        '1.2.4-beta.2',
         'x64'
       );
       await io.mkdirP(pythonDir);
       fs.writeFileSync(`${pythonDir}.complete`, 'hello');
     });
     // This will throw if it doesn't find it in the manifest (because no such version exists)
-    await finder.useCpythonVersion('1.2.3-beta.2', 'x64', false, false);
+    await expect(
+      finder.useCpythonVersion('1.2.4-beta.2', 'x64', false, false, false)
+    ).resolves.toEqual({
+      impl: 'CPython',
+      version: '1.2.4-beta.2'
+    });
   });
 
   it('Check-latest true, finds the latest version in the manifest', async () => {
@@ -176,7 +186,7 @@ describe('Finder tests', () => {
 
     fs.writeFileSync(`${pythonDir}.complete`, 'hello');
     // This will throw if it doesn't find it in the cache and in the manifest (because no such version exists)
-    await finder.useCpythonVersion('1.2', 'x64', true, true);
+    await finder.useCpythonVersion('1.2', 'x64', true, true, false);
 
     expect(infoSpy).toHaveBeenCalledWith("Resolved as '1.2.3'");
     expect(infoSpy).toHaveBeenCalledWith(
@@ -187,7 +197,7 @@ describe('Finder tests', () => {
     );
     expect(installSpy).toHaveBeenCalled();
     expect(addPathSpy).toHaveBeenCalledWith(expPath);
-    await finder.useCpythonVersion('1.2.3-beta.2', 'x64', false, true);
+    await finder.useCpythonVersion('1.2.4-beta.2', 'x64', false, true, false);
     expect(spyCoreAddPath).toHaveBeenCalled();
     expect(spyCoreExportVariable).toHaveBeenCalledWith(
       'pythonLocation',
@@ -199,11 +209,67 @@ describe('Finder tests', () => {
     );
   });
 
+  it('Finds stable Python version if it is not installed, but exists in the manifest, skipping newer pre-release', async () => {
+    const findSpy: jest.SpyInstance = jest.spyOn(tc, 'getManifestFromRepo');
+    findSpy.mockImplementation(() => <tc.IToolRelease[]>manifestData);
+
+    const installSpy: jest.SpyInstance = jest.spyOn(
+      installer,
+      'installCpythonFromRelease'
+    );
+    installSpy.mockImplementation(async () => {
+      const pythonDir: string = path.join(toolDir, 'Python', '1.2.3', 'x64');
+      await io.mkdirP(pythonDir);
+      fs.writeFileSync(`${pythonDir}.complete`, 'hello');
+    });
+    // This will throw if it doesn't find it in the cache and in the manifest (because no such version exists)
+    await expect(
+      finder.useCpythonVersion('1.2', 'x64', false, false, false)
+    ).resolves.toEqual({
+      impl: 'CPython',
+      version: '1.2.3'
+    });
+  });
+
+  it('Finds Python version if it is not installed, but exists in the manifest, pre-release fallback', async () => {
+    const findSpy: jest.SpyInstance = jest.spyOn(tc, 'getManifestFromRepo');
+    findSpy.mockImplementation(() => <tc.IToolRelease[]>manifestData);
+
+    const installSpy: jest.SpyInstance = jest.spyOn(
+      installer,
+      'installCpythonFromRelease'
+    );
+    installSpy.mockImplementation(async () => {
+      const pythonDir: string = path.join(
+        toolDir,
+        'Python',
+        '1.1.0-beta.2',
+        'x64'
+      );
+      await io.mkdirP(pythonDir);
+      fs.writeFileSync(`${pythonDir}.complete`, 'hello');
+    });
+    // This will throw if it doesn't find it in the cache and in the manifest (because no such version exists)
+    await expect(
+      finder.useCpythonVersion('1.1', 'x64', false, false, false)
+    ).rejects.toThrowError();
+    await expect(
+      finder.useCpythonVersion('1.1', 'x64', false, false, true)
+    ).resolves.toEqual({
+      impl: 'CPython',
+      version: '1.1.0-beta.2'
+    });
+    // Check 1.1.0 version specifier does not fallback to '1.1.0-beta.2'
+    await expect(
+      finder.useCpythonVersion('1.1.0', 'x64', false, false, true)
+    ).rejects.toThrowError();
+  });
+
   it('Errors if Python is not installed', async () => {
     // This will throw if it doesn't find it in the cache and in the manifest (because no such version exists)
     let thrown = false;
     try {
-      await finder.useCpythonVersion('3.300000', 'x64', true, false);
+      await finder.useCpythonVersion('3.300000', 'x64', true, false, false);
     } catch {
       thrown = true;
     }
--- a/action.yml
+++ b/action.yml
@@ -23,6 +23,9 @@ inputs:
   update-environment:
     description: "Set this option if you want the action to update environment variables."
     default: true
+  allow-prereleases:
+    description: "When 'true', a version range passed to 'python-version' input will match prerelease versions if no GA versions are found. Only 'x.y' version range is supported for CPython."
+    default: false
 outputs:
   python-version:
     description: "The installed Python or PyPy version. Useful when given a version range as input."
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -149,6 +149,35 @@ jobs:
     - name: Run simple code
       run: python -c 'import math; print(math.factorial(5))'
 
+  setup-prerelease-version:
+    name: Setup 3.12 ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-latest]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: setup-python 3.12
+      id: setup-python
+      uses: ./
+      with:
+        python-version: '3.12'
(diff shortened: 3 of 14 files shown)


dmitry-shibanov:
Hello @mayeut. Thank you for your pull request. I think the same behaviour can be achieved through specifying python-version input as `~3.10.0.alpha.x` and adding the [check-latest](https://github.com/actions/setup-python/pull/406) functionality. With this input for python-version it will handle unstable and stable versions.

mayeut:
@dmitry-shibanov,

> Thank you for your pull request. I think the same behaviour can be achieved through specifying python-version input as ~3.10.0.alpha.x and adding the https://github.com/actions/setup-python/pull/406 functionality. With this input for python-version it will handle unstable and stable versions.

The syntax proposed will still require workarounds and bash string manipulation when using `python-version` in a matrix together with tools like `nox`/`tox` (c.f. . https://github.com/pypa/packaging/pull/551 for example).

It's also not as easy as users expect it to be (https://github.com/actions/virtual-environments/issues/5540 / https://github.com/actions/virtual-environments/issues/3343).

Furthermore, check-lastest [will always download & install from manifest](https://github.com/actions/setup-python/blob/86e88a60ab96ff52fdfd576b37d772e4eda25f19/README.md?plain=1#L231-L233) rather than using tool-cache on every stable python update (I'd say for about a week once a month, time for GitHub-hosted runners to get the new python & be deployed) which is a waste of time & bandwidth that's avoided with this PR.
If you don't want to pay this price, then we're back to modifying the workflow twice in order to remove check-latest.

brcrista:
Thanks for the suggestion @mayeut.  I understand that this change would make certain use cases more ergonomic.  However, the amount of friction here is by design.

Actions uses NPM-style version patterns.  Here's why it [doesn't consider `3.11` a match for `3.11-alpha`](https://www.npmjs.com/package/semver):

> prerelease versions frequently are updated very quickly, and contain many breaking changes that are (by the author's design) not yet fit for public consumption. Therefore, by default, they are excluded from range matching semantics.

So while it's a good fit for the use case of wanting to test a package against a pre-release version, it opens up the opposite problem of someone needing to use special syntax to make sure their workflow is running with a GA version.  I think we should follow NPM's lead of defaulting to the GA case and requiring you to specify otherwise, rather than defaulting to the prerelease and requiring you to specify GA.

ssbarnea:
I am not sure who is the product manager on GHA, but I hope that is not the same person as the one saying "the amount of friction here is by design.".

The reality is that anyone wanting to test with multiple versions of python (or other tools) will want to be able to start building and testing with it before that version reaches GM. 

I would be really worries to find-out that, by design, GHA team makes the life harder for anyone trying to do this, forcing them to use more complex versioning or forcing them to have to change their pipelines the day one version of python reaches GM.

My hopes were that GHA team to continue working on making its use easier for most users, not harder. I know that sometimes historical mistakes may slow down the progress, but not wanting to provide the best possible experience is a big red flag.

It is bit hilarious to see someone from outside trying to help the team and write a pull-request, just to be refused with an answer like that.

brcrista:
@ssbarnea your feedback is welcome, but please [make your points respectfully and avoid criticizing anyone personally](https://github.com/actions/setup-python/blob/main/CODE_OF_CONDUCT.md#our-standards).  I think the point I made about consistency with the NPM version spec syntax and the other `setup-` actions still stands.

brcrista:
@mayeut I've been giving this some more thought.  In `setup-node`, there's a [`latest` keyword you can use](https://github.com/actions/setup-node#supported-version-syntax) that does what it says on the tin.  Since I imagine this fallback should only apply to the latest major/minor version, would that be a good fit for this use case?

webknjaz:
> I think the point I made about consistency with the NPM version spec syntax and the other `setup-` actions still stands.

OTOH, it's extremely weird to see an action for managing Python, that is expected to be used by pythonistas following versioning restrictions that are absolutely foreign to the Python ecosystem.
It's like coming to a restaurant and being handed chopsticks for eating a soup, just because the chef is also making a lot of sushi and needs consistency across all the meals.

mayeut:
> Since I imagine this fallback should only apply to the latest major/minor version, would that be a good fit for this use case?

@bcrista, no it doesn't, the purpose of this PR is to simplify workflows & interactions between the different tools using one version specifier consistently. I think most python library repos use a matrix of python version to test.

Consider the following workflow (& I'll focus on CPython because [PyPy is an entirely different  animal](https://github.com/actions/setup-python/issues/508)):
```yaml
jobs:
  test:
    name: ${{ matrix.os }} / ${{ matrix.python_version }}
    runs-on: ${{ matrix.os }}-latest
    strategy:
      fail-fast: false
      matrix:
        os: [Ubuntu, Windows, macOS]
        python_version: ["3.7", "3.8", "3.9", "3.10", "pypy3.7", "pypy3.8", "pypy3.9"]

    steps:
      - uses: actions/checkout@v3

      - uses: actions/setup-python@v4
        name: Install Python ${{ matrix.python_version }}
        with:
          python-version: ${{ matrix.python_version }}

      - name: Run nox
        run: pipx run nox --error-on-missing-interpreters -s tests-${{ matrix.python_version }}
```

Adding 3.11 is unnecessarily complex, here's one example:
```diff
diff --git a/.github/workflows/test.yml b/.github/workflows/test.yml
index 785c652..6490779 100644
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [Ubuntu, Windows, macOS]
-        python_version: ["3.7", "3.8", "3.9", "3.10", "pypy3.7", "pypy3.8", "pypy3.9"]
+        python_version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev", "pypy3.7", "pypy3.8", "pypy3.9"]
 
     steps:
       - uses: actions/checkout@v3
@@ -21,4 +21,9 @@ jobs:
           python-version: ${{ matrix.python_version }}
 
       - name: Run nox
-        run: pipx run nox --error-on-missing-interpreters -s tests-${{ matrix.python_version }}
+        run: |
+          # Need to remove "-dev" suffix
+          INTERPRETER=${{ matrix.python_version }}
+          INTERPRETER=${INTERPRETER/-dev/}
+          pipx run nox --error-on-missing-interpreters -s tests-${INTERPRETER}
+        shell: bash
```

First, you have to know how to implement the bash trick.
Second, it's not really what you want to do, you don't really want 3.11-dev which could install 3.11.1a1 later on ([it's not the case at the moment, but I would expect it to be the case](https://github.com/actions/setup-python/issues/416#issuecomment-1150366180)) so you'll need to revert this one to 3.11 later on.

This could also be done slightly differently:
```diff
diff --git a/.github/workflows/test.yml b/.github/workflows/test.yml
index 785c652..1e00652 100644
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [Ubuntu, Windows, macOS]
-        python_version: ["3.7", "3.8", "3.9", "3.10", "pypy3.7", "pypy3.8", "pypy3.9"]
+        python_version: ["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.7", "pypy3.8", "pypy3.9"]
 
     steps:
       - uses: actions/checkout@v3
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/setup-python@v4
         name: Install Python ${{ matrix.python_version }}
         with:
-          python-version: ${{ matrix.python_version }}
+          python-version: "${{ startsWith(matrix.python_version, 'pypy') && matrix.python_version || format('~{0}.0-0', matrix.python_version) }}"
 
       - name: Run nox
         run: pipx run nox --error-on-missing-interpreters -s tests-${{ matrix.python_version }}
```
This one requires knowledge of GHA expressions (let's say intermediate level) and NPM version spec syntax and as @webknjaz said, the intended audience for this action is not the NPM users but Python users. Even if it were, the wrong answer was proposed by a maintainer meaning it's not as trivial as it sounds.

There are for sure other ways to do that but I'm almost certain they involve to do something more complex than should be.

> However, the amount of friction here is by design.

IMHO, `X.Y` & `pypyX.Y` should be considered as special citizens here & not involve that amount of friction. Since I wrote this PR, #213 has been closed and now, another issue has been created: #501.
I understand the whole generic idea behind the spec to only match GA version as a default but we're not in a generic case here. The user explicitly requested an `X.Y` version for which no GA exists yet, we should install  the latest pre-release in this case.
I could rework this with another input to the action which would then still be easier e.g.
```diff
diff --git a/.github/workflows/test.yml b/.github/workflows/test.yml
index 785c652..1e00652 100644
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [Ubuntu, Windows, macOS]
-        python_version: ["3.7", "3.8", "3.9", "3.10", "pypy3.7", "pypy3.8", "pypy3.9"]
+        python_version: ["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.7", "pypy3.8", "pypy3.9"]
 
     steps:
       - uses: actions/checkout@v3
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/setup-python@v4
         name: Install Python ${{ matrix.python_version }}
         with:
           python-version: ${{ matrix.python_version }}
+          fallback-no-ga: "allow"  # could be "forbid" (default) or "warn" ?
 
       - name: Run nox
         run: pipx run nox --error-on-missing-interpreters -s tests-${{ matrix.python_version }}
```

EwoutH:
I would just like to say that I support enabling to use preview Python versions without the `-dev` suffix, since it will reduce maintenance efforts. Thanks for the PR!

aaugustin:
I would also enjoy the ability to put "3.11" in my GHA config when it's still a prerelease and forgetting about it :-)

The friction of having to set "3.11-dev" and then having to remember to come back to change it to "3.11" isn't improving my life as a GHA user. Not to mention the added friction of having to explain the situation to well-meaning contributors who aren't aware of these contraints.

In addition, I'd like to minimize the amount of time I spend dealing with version updates. It isn't the most valuable use of my open-source maintenance time, to say the least.

I understand that the proposed improvement here changes the behavior for a user who doesn't want to test on a given version before it's GA and accidentally configures it anyway. Instead of "crash - Python version not found", the behavior becomes "run on a pre-release". I would argue that a user who configures a given Python version probably wants to run that version. I wouldn't assume that they want to crash if that version is still in prerelease.

I read the comments about creating friction on purpose but I'm not convinced by the benefits. If one doesn't want to test on a given Python version before it's GA, then one has no reason to add it in the GHA config until it's GA anyway. It looks like the current design degrades the experience of the majority to cater for the minority that makes a mistake by adding new Python versions earlier than intended, before GA. This doesn't sound like a good tradeoff.

randydwhitephotography (approved):
installer build-extra

hugovk:
It would be great to avoid having to repeat this next year for `3.12-dev` -> `3.12`:

Big list of PRs just to replace 3.11-dev with 3.11

webknjaz:
> It would be great to avoid having to repeat this next year for 3.12-dev -> 3.12 (and I'm not finished yet):

@hugovk I was reflecting on this for a while and realized that similar updates would probably still be necessary. Just think about it: usually, it's unstable at the beginning. So it's typical to allow failures for the dev version. With `~` or `-dev`, it's currently possible to use these markers in the condition for `continue-on-failure`. If it's going to be just the version, one would have to introduce an experimental flag in the matrix. And as the version becomes stable, they'd still have to make a change to drop that flag.
So strictly speaking, not having to update the matrix factor in one place would just shift the same activity to another matrix factor and that's it.

hugovk:
I don't know, none of my ~25 PRs above needed that, `3.11-dev` was just a normal matrix job 🤷

aaugustin:
I don't need a two step process either. I can make one PR that says "3.12" (when it's in beta or RC) and makes required changes (there are usually few or none). No need for more!

brcrista:
@mayeut to move forward with this, let's add it as an optional input: `allow-prereleases: true`.

Rationale:
1. Not matching prereleases by default follows SemVer and is how [Pip](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-prereleases) and [Poetry](https://python-poetry.org/docs/dependency-specification/#expanded-dependency-specification-syntax) work.  It's also consistent with the rest of the `setup-` actions.
2. It avoids a breaking change to `setup-python`.

Thoughts?

webknjaz:
> I don't know, none of my ~25 PRs above needed that, 3.11-dev was just a normal matrix job 🤷

@hugovk how many of those are Cython-based? For pure Python the problem described is indeed not as evident but if a project depends on a build tool that needs to make compatibility changes because CPython's ABI got updated, it adds a substantial amount of delay in the early stages of the release cycle. This is why we don't test against Cython stuff against the unreleased Python versions in @aio-libs, for example. This is the case I was thinking about.

mayeut:
@brcrista, I updated the PR with an optional input: `allow-prereleases: true`. I still need to update the docs with the new input.

action.yml
```diff
@@ -23,6 +23,9 @@ inputs:
   update-environment:
     description: "Set this option if you want the action to update environment variables."
     default: true
+  allow-prereleases:
+    description: "Set this option if you want the action to allow installing pre-releases python versions when no GA version matches."
```
brcrista, line 27:
```suggestion
    description: "When 'true', a version range passed as to 'python-version' input will match prerelease versions if no GA versions are found."
```

__tests__/finder.test.ts
```diff
@@ -199,11 +209,63 @@ describe('Finder tests', () => {
...
+    // This will throw if it doesn't find it in the cache and in the manifest (because no such version exists)
+    await expect(
+      finder.useCpythonVersion('1.2', 'x64', false, false, false)
+    ).resolves.toEqual({
+      impl: 'CPython',
+      version: '1.2.3'
+    });
+  });
+
+  it('Finds Python version if it is not installed, but exists in the manifest, pre-release fallback', async () => {
```
brcrista, line 234: There are two kinds of version inputs you can pass: an exact version and a "version range."
- An exact version is something like `3.11.0`
- The syntax and matching logic for version ranges is defined [here](https://www.npmjs.com/package/semver)

So, the version range `1.1` means "give me the latest sub-version of 1.1" and matches either `1.1.0` or `1.1.0-beta.2` depending on the circumstances.  However, I would expect the exact version `1.1.0` _not_ to match `1.1.0-beta.2`, even if `allow-prerelease` is set.  On the other hand, `>=1.1.0` is a version range and _would_ match if a GA `1.1.0` isn't available.

Could you add tests for those cases?

e-korolevskii (approved):

__tests__/finder.test.ts
mayeut: I updated the docs, for CPython, only `x.y` version range is supported, hence no need to add test cases for other ranges.

brcrista: > For CPython, `allow-prereleases` will only have effect for `x.y` version range (e.g. `3.12`).

Can you add a test that confirms the behavior for when you pass `allow-prerelease: true` for a non-`x.y` version?  I'm assuming it will ignore it -- maybe we should print a warning as well?

Anyway, having a test case will make sure it doesn't break in a weird way, and make sure it doesn't accidentally start doing something unintended that people come to depend on.

mayeut: I added a test for 1.1.0 version specifier not falling back to 1.1.0-beta.2

mayeut:
@brcrista, I think I answered all remarks. Anything I missed to get this merged ?

brcrista (approved):
Thanks for the contribution @mayeut !

hugovk:
Thank you very much for this PR!

It would be great if we can start using it, do you have an idea when it will be released?

PR #415, opened 2022-06-05
Updates to setup-python this week

codenamedpkt:
**Description:**
<br>For the setup-python repo by official GitHub @actions, I update the OS requirements to the latest @Ubuntu 22.04. Lot of Action files has modified or trimmed due to the recent @Python official's announcements.</br>

**Related issue:**
<br>No issues related to these changes.</br>

**Check list:**


marko-zivic-93:
Hello @codenamedpkt 
Could you please provide more detailed justification for this PR? What benefits do we expect from these changes?

.github/workflows/check-dist.yml
```diff
@@ -18,15 +18,15 @@ on:
 
 jobs:
   check-dist:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
Kurt-von-Laven, line 22: Ceasing to pin all of these versions (particularly all at once) seems very risky especially considering how many different teams rely on the setup-python action in so many different systems. If anything goes wrong, debugging a change this sweeping would generate a lot of unnecessary (and highly stressful) work for the maintainers.

.github/pull_request.md
```diff
@@ -0,0 +1,9 @@
+**Description:**
```
Kurt-von-Laven, line 1: The details of one pull request are not an appropriate pull request template for all future pull requests.

action.yml
```diff
@@ -25,7 +25,7 @@ outputs:
   python-path:
     description: "The absolute path to the Python executable."
 runs:
-  using: 'node16'
+  using: 'node18'
```
Kurt-von-Laven, line 28: `node18` isn't even the LTS version of Node.js currently recommended for most users yet. This seems premature.

yaml-lint-config.yml
```diff
@@ -1,7 +1,7 @@
 extends: default
 
 rules:
-  # 120 chars should be enough, but don't fail if a line is longer
+  # 600 chars should be enough, but don't fail if a line is longer
```
Kurt-von-Laven, line 4: 600 characters is an extremely long line length and necessitates a great deal of tedious side-scrolling. 120 is the longest line length I have commonly seen enforced, and there is research suggesting that the optimal line length for readability is somewhere in the realm of 50-90 characters (at least in English).

Kurt-von-Laven (changes requested):
I am not affiliated with the setup-python action, but I feel this pull request should be closed for its overly broad scope and level of risk. I recommend that smaller more focused pull requests with more specific motivations clarified to achieve the intended goals of the author.

dmitry-shibanov:
Hello @codenamedpkt. We're not planing to accept your pull request because 

1. We do not recommend to use the main branch as a version tag for the action because it can contain breaking or untested changes. It's better to stick with major tag.
2. We'd like to use node 16 for setup-node because it's current LTS version and the setup-python action runs on node16
3. We'd like to use ubuntu-latest and not to stick with some specific os version.
4. We're not planing for now to change action's node version. Besides, Node 18 is not LTS yet.

yaml-lint-config.yml
Kurt-von-Laven: Eliminating warnings by permitting sloppy code seems counterproductive.

action.yml
Kurt-von-Laven: Stability and security are far greater concerns here. As far as I am aware, this particular codebase currently has no use for the latest Node.js features.

.github/workflows/check-dist.yml
Kurt-von-Laven: It's not merely the beta OS you are proposing using, but also the latest version of many other tools. I would pause to reflect on why almost no competent software companies choose to use the beta version of every tool they rely on when plenty of automation exists that would allow them to do so if they so desired.

yaml-lint-config.yml
Kurt-von-Laven: Sorry, but I don't understand the question.

Kurt-von-Laven: yamllint has [no such feature](https://yamllint.readthedocs.io/en/stable/rules.html), but even if it did, I don't see what relevance that would have to this discussion. Regardless of the units, it is ill-advised to raise the line limit to the equivalent of roughly 5 times the width of a typical monitor at a standard font size. One of the purposes of linters like yamllint is to prevent people from writing code in a way that most people would prefer not to read, and scrolling side-to-side is known to reduce the speed with which people can read.

Issue #416, opened 2022-06-05
bug: using `python-version: 3.9-dev` installs `3.10.4`


mayeut:
**Description:**

using `python-version: 3.9-dev` installs `3.10.4`

```
Run ./
  with:
    python-version: 3.9-dev
    token: ***
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.7.13/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.7.13/x64/lib/pkgconfig
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.7.13/x64/lib
Successfully set up CPython (3.10.4)
```

**Action version:**
v3, main

**Platform:**
- [x] Ubuntu
- [x] macOS
- [x] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**
`3.9-dev`

**Repro steps:**  
c.f description

**Expected behavior:**
Set up the most recent CPython 3.9 version available including pre-releases i.e. in [node-semver](https://github.com/npm/node-semver), `~3.9.0-0, {includePrerelease: true}`

**Actual behavior:**
Set up CPython 3.10.4

panticmilos:
Hi @mayeut,

I was able to reproduce this issue, we will take look at it and related PRs today.

dmitry-shibanov:
Hello @mayeut. The new major version was released with your fix. Could you please confirm that everything works expected?

mayeut:
@dmitry-shibanov,
The major issue is fixed (the one from the title).
I wouldn't say "as expected": it wouldn't have installed 3.9.2rc1 back in time, it would have used a 3.9.0 or 3.9.1 instead.
However given the following, I will let you decide if you want to rename the issue or just close it:
- there are very few pre-releases from cpython x.y once a x.y.0 has been released (the last one was [3.9.2rc1](https://github.com/python/cpython/releases/tag/v3.9.2rc1) on 16 Feb 2021)
- would require #406 (but switched on automatically by the `-dev` suffix)
- [an issue in the actions/tool-cache package](https://github.com/actions/toolkit/issues/709) (In the end, I'm not sure if this one is relevant if #406 is there)

dmitry-shibanov:
Hello @mayeut. It looks like listed pull requests were merged. Now it works as expected. For now I'm going to close the issue if you have any concerns feel free to ping us.

PR #417, opened 2022-06-05
fix: 3.x-dev can install a 3.y version

mayeut:
**Description:**
Rework `desugarDevVersion` so that `3.x-dev` cannot end up with `3.y`/`4.x` installed.
It still does not do what one would expect (include pre-release even if there's a stable release - would require https://github.com/actions/toolkit/issues/709 to be fixed) but it was already the case & [python-versions](https://github.com/actions/python-versions) does not build pre-releases for minor versions once a stable release exists, so it does not matter much (but going with mitigates rather than fix below).

**Related issue:**
mitigates #416 

**Check list:**

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -1,5 +1,5 @@
 name: Validate Python e2e
-on: 
+on:
   push:
     branches:
       - main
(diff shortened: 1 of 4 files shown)


panticmilos (approved):

vsafonkin:
Hi @mayeut, could you please update [README](https://github.com/actions/setup-python/blob/main/README.md#available-versions-of-python) note about `-dev` syntax?

```md
- Use `-dev` instead of a patch number (e.g., `3.11-dev`) to install the latest release of a patch version, *alpha and beta releases included
```

marko-zivic-93 (approved):

mayeut:
> could you please update [README](https://github.com/actions/setup-python/blob/main/README.md#available-versions-of-python) note about -dev syntax?

@vsafonkin,
what should it be changed to ? the implementation is just being fixed, should not change the doc. unless you want to account for the fact that alpha and beta are not included once a stable version exists in which case:
```
- Use `-dev` instead of a patch number (e.g., `3.11-dev`) to install the latest release of a patch version, *alpha and beta releases included if no stable version exists*
```

vsafonkin:
@mayeut, I mean `to install the latest release of a minor version` -> `to install the latest release of a patch version` replacing

mayeut:
It was clear for me with `minor`, I find it unclear with your proposal. How about:
`to install the latest release of a minor version` -> `to install the latest patch version release for a given minor version`

If what the patch is doing is unclear:
```
asking "x.y-dev"
["x.y.0-a"] -> "x.y.0-a"
["x.y.0-a", "x.y.0"] -> "x.y.0"
["x.y.0-a", "x.(y+1).0"] -> "x.y.0-a"
["x.y.0-a", "x.y.1"] -> "x.y.1"
["x.y.0-a", "x.y.1-a"] -> "x.y.0-a"
```

marko-zivic-93 (approved):

PR #427, opened 2022-06-10
Pass the `token` input through on GHES

brcrista:
Closes https://github.com/actions/setup-python/issues/229

## Description
- Customers hit this issue because they have a fleet of self-hosted runners behind a handful of NAT gateways.  So, all of those runners look like the same small set of IPs to GitHub.
- We currently [null out the PAT ](https://github.com/actions/setup-python/blob/main/src/install-python.ts#L9)when we're on GHES.  This was done just to avoid confusion and not for any security reason.
- Passing a github.com PAT to the `token` input is an inexpensive fix that gets the job done without a lot of complexity.  While GitHub Connect is the ideal pattern here, using `token` doesn't seem that bad in this case.  The [description](https://github.com/actions/setup-python/blob/813f9b1556d9f0a0faf7b08a19486d6af9857537/action.yml#L16) for the token input says:
  > Used to pull python distributions from actions/python-versions
  
  Users already seem to expect it to be a [github.com](http://github.com/) PAT, and being dropped silently is already unintuitive.

Given that we're not going to fund the GitHub Connect work, I think just **removing the code to null out the PAT on GHES and updating the description for `token` to make the behavior clear** is the best fix.

## Testing

1. Pushed my version of the action to a GHES instance and configured a self-hosted runner
2. Created a PAT for github.com (no scopes) and added it as a secret
3. Ran the YAML snippet from the issue
4. Confirmed that there was no change to the unauthenticated rate limit

## Check list


dmitry-shibanov (approved):

panticmilos (approved):

brcrista:
The failing checks are for `pipenv` caching on Windows.  Doesn't appear to be related, and appears in runs for other branches.

PR #431, opened 2022-06-12
Allow `python-version-file` to be a relative path

### LINKED ISSUE #428: Log statement omits default Python version file
**Description:**
When neither `python-version` nor `python-version-file` is specified, the `setup-python` correctly defaults to reading the Python version from `.python-version` as documented. However, it logs: `Resolved  as x.y.z`.

**Action version:**
v4.0.0

**Platform:**
- [x] Ubuntu
- [x] macOS
- [x] Windows

**Runner type:**
- [x] Hosted
- [x] Self-hosted

**Tools version:**
All

**Repro steps:**

1. Create a `.python-version` file containing `3.10.4`.
1. Add the following to your workflow:
    
    ```yaml
    
    - name: Set up Python based on .python-version file.
      uses: actions/setup-python@v4.0.0
    ```
    
1. Run your workflow.
1. Examine the logs.

**Expected behavior:**
I expected the logs to include: `Resolved .python-version as 3.10.4`.

**Actual behavior:**
The pertinent log statement says: `Resolved  as 3.10.4`.

### LINKED ISSUE #430: Can't Specify `python-version-file` as Absolute Path
**Description:**
The `python-version-file` parameter currently only functions correctly with relative paths.

**Action version:**
v4.0.0

**Platform:**
- [x] Ubuntu
- [x] macOS
- [x] Windows

**Runner type:**
- [x] Hosted
- [x] Self-hosted

**Tools version:**
All

**Repro steps:**
See ScribeMD/slack-templates@e653014e1fcbd1284b8cff3fe0a6c9fc735793ca and [its accompanying workflow run](https://github.com/ScribeMD/slack-templates/actions/runs/2481683517) for a reproduction.

1. Create a composite action.
1. Create a `.python-version` file containing `3.10.4`.
1. Call the setup-python action using an absolute path for the `python-version-file`:

    ```yaml
    - name: Set up Python based on .python-version file.
      uses: actions/setup-python@v4.0.0
      with:
        python-version-file: "${{ github.action_path }}/.python-version"
    ```

**Expected behavior:**
I expected Python 3.10.4 to be set up.

**Actual behavior:**
The setup-python action crashes with:
`Error: The specified python version file at: /home/runner/work/<action_name>/<action_name>/home/runner/work/<action_name>/<action_name>/.python-version does not exist`

Kurt-von-Laven:
**Description:**
Don't assume that it is safe to prepend the `GITHUB_WORKSPACE` environment variable to the given path since the path may already be absolute.

**Related issue:**
Fixes #428.
Fixes #430.

**Check list:**

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -64546,18 +64546,18 @@ function cacheDependencies(cache, pythonVersion) {
 }
 function resolveVersionInput() {
     let version = core.getInput('python-version');
-    const versionFile = core.getInput('python-version-file');
+    let versionFile = core.getInput('python-version-file');
     if (version && versionFile) {
         core.warning('Both python-version and python-version-file inputs are specified, only python-version will be used');
     }
     if (version) {
         return version;
     }
-    const versionFilePath = path.join(process.env.GITHUB_WORKSPACE, versionFile || '.python-version');
-    if (!fs_1.default.existsSync(versionFilePath)) {
-        throw new Error(`The specified python version file at: ${versionFilePath} does not exist`);
+    versionFile = versionFile || '.python-version';
+    if (!fs_1.default.existsSync(versionFile)) {
+        throw new Error(`The specified python version file at: ${versionFile} does not exist`);
     }
-    version = fs_1.default.readFileSync(versionFilePath, 'utf8');
+    version = fs_1.default.readFileSync(versionFile, 'utf8');
     core.info(`Resolved ${versionFile} as ${version}`);
     return version;
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "setup-python",
-  "version": "3.1.1",
+  "version": "4.0.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "setup-python",
-      "version": "3.1.1",
+      "version": "4.0.0",
       "license": "MIT",
       "dependencies": {
         "@actions/cache": "^2.0.2",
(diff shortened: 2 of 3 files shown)


vsafonkin:
@Kurt-von-Laven, probably it will be better to add a check for `python-version-file` input is absolute or relative:
https://nodejs.org/api/path.html#pathisabsolutepath

And skip joining with `GITHUB_WORKSPACE` var if a path from this input is absolute. What do you think about it?

Kurt-von-Laven:
I can’t think of a benefit the extra branch would confer, but I certainly don’t claim comprehensive knowledge of GitHub Actions. My understanding is that `GITHUB_WORKSPACE` is already the default working directory of steps. It’s nice to be able to override this default by setting the `working-directory` explicitly. Notably, in composite actions `GITHUB_WORKSPACE` is usually the wrong default working directory in my experience.

Kurt-von-Laven:
We are saying the same thing. The default working directory when not specified by explicitly setting `working-directory` is always `GITHUB_WORKSPACE`. For this reason alone, it seems odd to explicitly prepend `GITHUB_WORKSPACE` even if only to relative paths, because as far as I can see doing so never adds any value. The bigger problem though is that the action effectively hardcodes the working directory, causing the user's explicitly specified choice of `working-directory` to be silently ignored. This is most problematic in my experience within composite actions, where the default working directory of `GITHUB_WORKSPACE` frequently needs to be overridden to instead be the directory the composite action was checked out to.

dmitry-shibanov:
Hello @Kurt-von-Laven. Could you please sync with the main branch ?

dmitry-shibanov (approved):

marko-zivic-93 (approved):

Kurt-von-Laven:
Thank you for the review everyone! I rebased on main.

PR #434, opened 2022-06-14
add restore cache error handling

panticmilos:
**Description:**
This PR provides error handling for cache restore functionality.

**Check list:**

### DIFF
--- a/src/cache-distributions/cache-distributor.ts
+++ b/src/cache-distributions/cache-distributor.ts
@@ -35,13 +35,25 @@ abstract class CacheDistributor {
     core.saveState(State.CACHE_PATHS, cachePath);
     core.saveState(State.STATE_CACHE_PRIMARY_KEY, primaryKey);
 
-    const matchedKey = await cache.restoreCache(
-      cachePath,
-      primaryKey,
-      restoreKey
-    );
+    try {
+      const matchedKey = await cache.restoreCache(
+        cachePath,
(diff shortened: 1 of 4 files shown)


marko-zivic-93 (approved):

dmitry-shibanov (approved):

vsafonkin (approved):

src/cache-distributions/cache-distributor.ts
```diff
@@ -35,13 +35,23 @@ abstract class CacheDistributor {
     core.saveState(State.CACHE_PATHS, cachePath);
     core.saveState(State.STATE_CACHE_PRIMARY_KEY, primaryKey);
 
-    const matchedKey = await cache.restoreCache(
-      cachePath,
-      primaryKey,
-      restoreKey
-    );
+    try {
+      const matchedKey = await cache.restoreCache(
```
brcrista, line 39: Shouldn't we push this fix into [`restoreCache` from `actions/cache`](https://github.com/actions/toolkit/blob/b5f31bb5a25d129441c294fc81ba7f92f3e978ba/packages/cache/src/cache.ts#L65)?  That way everything that uses caching will handle 503s correctly.

brcrista (changes requested):
I think we should push the fix into the `actions/cache` package so everything gets fixed and is future-proofed

PR #437, opened 2022-06-15
Revert "Pass the `token` input through on GHES"

brcrista:
Reverts actions/setup-python#427

@thboop pointed out a bug in this: if you _don't_ pass a token and it uses the default `github.token`, then pulling a new version that isn't in the tool cache will fail.  The GHES `github.token` is not valid for calling github.com.


thboop (approved):
lgtm

marko-zivic-93 (approved):

vsafonkin (approved):

Issue #231, opened 2021-08-10
Request PYTHON_INCLUDE_DIR and PYTHON_LIBRARY variables (or equivalent)


ramcdona:
CMake's FindPythonLibs.cmake searches standard install locations to find a Python installation to build against when compiling C/C++ to work with Python.  This generally means it finds the default system-installed Python.

Since setup-python installs specific requested versions of Python to a non-standard location, CMake won't ever find them.

It would be great if setup-python set variables (like pythonLocation) that point at the library and header files needed to build C/C++ against a specific Python.  In CMake terms, these are PYTHON_INCLUDE_DIR and PYTHON_LIBRARY, but other names could be used.

The CMake scripts FindPythonLibs, FindPythonInterp, and the new FindPython2 and FindPython3 set some other variables that might also be worth adding to the list.

Finally, it would be great if all these variables were clearly documented and their existence was obvious to a user of setup-python.

dsame:
hello @ramcdona , i [created the PR](https://github.com/actions/setup-python/pull/440) to set the hints FindPython, FindPython2 and FindPython3

I hesitate we should set PYTHON_INCLUDE_DIR and PYTHON_LIBRARY variables for deprecated FindPythonLibs module - setting these variables potentially can override the default search based on ROOT_DIR. Any thougts?

ramcdona:
I'm sorry, I thought I was included on this review by mistake.  As a GH Actions user (not developer), I don't know how to test this change in my build system to see if it resolves my problems.

Obviously, my goal is for FindPythonXXX to 'just work' when using setup-python to set a particular version.

CMake 3.12 (the version that deprecates FindPythonLibs) is pretty new in CMake world.  While the CMake community is getting better at using newer versions, it still lags.  [This site](https://gitlab.kitware.com/cmake/community/-/wikis/CMake-Versions-on-Linux-Distros) lists what CMake version can be expected by default with most Linux distributions.

The [GH Actions Ubuntu](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md) 18.04 has CMake 3.23.2, so that isn't a problem.  It looks like the oldest Windows and MacOS environments also support 3.22.2.

In my own program, I only recently updated my CMAKE_MINIMUM_REQUIRED to 3.1.  I have had need to support RH 7 -- which is terribly slow to progress.

[...]

PR #440, opened 2022-06-22
Add CMake hints

dsame:
**Description:**

FindPythonNNN CMake modules uses the environment variables to find the specific version of Python

The change is similar to setting `pythonLocation` and `PKG_CONFIG_PATH` env. variables.

**Related issue:**
https://github.com/actions/setup-python/issues/231


panticmilos (approved):

marko-zivic-93 (approved):

IvanZosimov (approved):

dsame:
> I'm not sure that we should add some custom environment variables for any tools to the action. There are many tools with a lot of environment variables to configure it, are we going to add more and more of them further? I think it should happen from the customers side.

cmake is very important tool and it is expected to work out of box. It works with default python installation without extra configuration, so we should to keep the same behaviour.

marko-zivic-93 (approved):

Issue #446, opened 2022-06-27
Wrong Poetry config is used when project is not in repository's root


oranav:
**Description:**
When collecting Poetry projects for caching, a `**/poetry.lock` glob is used, meaning the action supports Poetry projects in subfolders.

<https://github.com/actions/setup-python/blob/ffcd00020c213885323d35524aa155f17870b8dd/src/cache-distributions/poetry-cache.ts#L11>

However, in order to process the Poetry configuration, `getPoetryConfiguration` runs the `poetry` command from the repo's root directory; this causes Poetry to return an invalid configuration when there is a Poetry project inside an inner directory.

<https://github.com/actions/setup-python/blob/ffcd00020c213885323d35524aa155f17870b8dd/src/cache-distributions/poetry-cache.ts#L45-L48>

**Action version:**
v4

**Platform:**
- [X] Ubuntu
- [X] macOS
- [X] Windows

**Runner type:**
- [X] Hosted
- [X] Self-hosted

**Tools version:**

All Poetry versions

**Repro steps:**  
Create a repository with a Poetry project inside a subfolder; use the default action configuration. Matter of fact, even if a specific  `cache-dependency-path` is provided, the resulting Poetry config will still be invalid.

**Expected behavior:**
The correct pyproject.toml and poetry.toml (inside the subfolder) are used.

**Actual behavior:**
The read configuration uses pyproject.toml and poetry.toml from the repository root directory.

IvanZosimov:
Hi, @oranav 👋 ! Thank you for this issue! We will investigate it and get back to you with updates 🖨️

oranav:
@IvanZosimov I already have a draft PR here: #447, I just have to test it locally first (and maybe add a test)

dsame:
Hello @oranav 

Can you please help to reproduce the problem. 
I tried to create the workflow according to the description and it looks like everything works as expected

This [build](https://github.com/akv-demo/setup-python-test/actions/runs/3359696666/jobs/5567997165#step:9:44) creates the cache, you can see it from `Post Run actions/setup-python@v4` step.

And [next run](https://github.com/akv-demo/setup-python-test/actions/runs/3359700718/jobs/5568006451#step:5:33) of the same pipeline definitely uses the cache from the previous step because of the output of  `run` says:

```
Installing dependencies from lock file
No dependencies to install or update
```

[...]

PR #447, opened 2022-06-27
Use correct Poetry config when collecting Poetry projects

oranav:
**Description:**
When collecting Poetry projects for caching, a `**/poetry.lock` glob is used.  However, in order to process the Poetry configuration, the "poetry" command is run from the repo's root directory; this causes Poetry to return an invalid configuration when there is a Poetry project inside an inner directory.

Instead of running a single Poetry command, glob for the same pattern, and configure Poetry for every discovered project.

**Related issue:**
#446 

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ The action defaults to searching for a dependency file (`requirements.txt` for p
 
  - For `pip`, the action will cache the global cache directory
  - For `pipenv`, the action will cache virtualenv directory
- - For `poetry`, the action will cache virtualenv directory
+ - For `poetry`, the action will cache virtualenv directories -- one for each poetry project found
 
 **Caching pip dependencies:**
 
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -27,7 +27,7 @@ virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/patrick/Library/Caches/py
   let infoSpy: jest.SpyInstance;
   let warningSpy: jest.SpyInstance;
   let debugSpy: jest.SpyInstance;
-  let saveSatetSpy: jest.SpyInstance;
+  let saveStateSpy: jest.SpyInstance;
   let getStateSpy: jest.SpyInstance;
   let setOutputSpy: jest.SpyInstance;
 
@@ -52,8 +52,8 @@ virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/patrick/Library/Caches/py
     debugSpy = jest.spyOn(core, 'debug');
     debugSpy.mockImplementation(input => undefined);
 
-    saveSatetSpy = jest.spyOn(core, 'saveState');
-    saveSatetSpy.mockImplementation(input => undefined);
+    saveStateSpy = jest.spyOn(core, 'saveState');
+    saveStateSpy.mockImplementation(input => undefined);
 
     getStateSpy = jest.spyOn(core, 'getState');
     getStateSpy.mockImplementation(input => undefined);
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -183,6 +183,10 @@ virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/patrick/Library/Caches/py
           expect(infoSpy).toHaveBeenCalledWith(
             `Cache restored from key: setup-python-${process.env['RUNNER_OS']}-20.04-Ubuntu-python-${pythonVersion}-${packageManager}-${fileHash}`
           );
+        } else if (packageManager === 'poetry') {
+          expect(infoSpy).toHaveBeenCalledWith(
+            `Cache restored from key: setup-python-${process.env['RUNNER_OS']}-python-${pythonVersion}-${packageManager}-v2-${fileHash}`
+          );
         } else {
           expect(infoSpy).toHaveBeenCalledWith(
             `Cache restored from key: setup-python-${process.env['RUNNER_OS']}-python-${pythonVersion}-${packageManager}-${fileHash}`
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -1,8 +1,10 @@
+import * as path from 'path';
 import * as core from '@actions/core';
 import * as cache from '@actions/cache';
 import * as exec from '@actions/exec';
 import * as io from '@actions/io';
 import {getCacheDistributor} from '../src/cache-distributions/cache-factory';
+import {State} from '../src/cache-distributions/cache-distributor';
 import * as utils from './../src/utils';
 
 describe('restore-cache', () => {
@@ -13,7 +15,7 @@ describe('restore-cache', () => {
   const requirementsLinuxHash =
     '2d0ff7f46b0e120e3d3294db65768b474934242637b9899b873e6283dfd16d7c';
   const poetryLockHash =
-    '571bf984f8d210e6a97f854e479fdd4a2b5af67b5fdac109ec337a0ea16e7836';
+    'f24ea1ad73968e6c8d80c16a093ade72d9332c433aeef979a0dd943e6a99b2ab';
   const poetryConfigOutput = `
 cache-dir = "/Users/patrick/Library/Caches/pypoetry"
 experimental.new-installer = false
@@ -100,21 +102,56 @@ virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/patrick/Library/Caches/py
 
   describe('Restore dependencies', () => {
     it.each([
-      ['pip', '3.8.12', undefined, requirementsHash],
-      ['pip', '3.8.12', '**/requirements-linux.txt', requirementsLinuxHash],
+      ['pip', '3.8.12', undefined, requirementsHash, undefined],
+      [
+        'pip',
+        '3.8.12',
+        '**/requirements-linux.txt',
+        requirementsLinuxHash,
+        undefined
+      ],
       [
         'pip',
         '3.8.12',
         '__tests__/data/requirements-linux.txt',
-        requirementsLinuxHash
+        requirementsLinuxHash,
+        undefined
       ],
-      ['pip', '3.8.12', '__tests__/data/requirements.txt', requirementsHash],
-      ['pipenv', '3.9.1', undefined, pipFileLockHash],
-      ['pipenv', '3.9.12', '__tests__/data/requirements.txt', requirementsHash],
-      ['poetry', '3.9.1', undefined, poetryLockHash]
+      [
+        'pip',
+        '3.8.12',
+        '__tests__/data/requirements.txt',
+        requirementsHash,
+        undefined
+      ],
+      ['pipenv', '3.9.1', undefined, pipFileLockHash, undefined],
+      [
+        'pipenv',
+        '3.9.12',
+        '__tests__/data/requirements.txt',
+        requirementsHash,
+        undefined
+      ],
+      [
+        'poetry',
+        '3.9.1',
+        undefined,
+        poetryLockHash,
+        [
+          '/Users/patrick/Library/Caches/pypoetry/virtualenvs',
+          path.join(__dirname, 'data', 'inner', '.venv'),
+          path.join(__dirname, 'data', '.venv')
+        ]
+      ]
     ])(
       'restored dependencies for %s by primaryKey',
-      async (packageManager, pythonVersion, dependencyFile, fileHash) => {
+      async (
+        packageManager,
+        pythonVersion,
+        dependencyFile,
+        fileHash,
+        cachePaths
+      ) => {
         const cacheDistributor = getCacheDistributor(
           packageManager,
           pythonVersion,
@@ -122,6 +159,13 @@ virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/patrick/Library/Caches/py
         );
 
         await cacheDistributor.restoreCache();
+
+        if (cachePaths !== undefined) {
+          expect(saveStateSpy).toHaveBeenCalledWith(
+            State.CACHE_PATHS,
+            cachePaths
+          );
+        }
 
         if (process.platform === 'linux' && packageManager === 'pip') {
           expect(infoSpy).toHaveBeenCalledWith(
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -102,11 +102,17 @@ virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/patrick/Library/Caches/py
 
   describe('Restore dependencies', () => {
     it.each([
-      ['pip', '3.8.12', undefined, requirementsHash, undefined],
       [
         'pip',
         '3.8.12',
-        '**/requirements-linux.txt',
+        '__tests__/data/**/requirements.txt',
+        requirementsHash,
+        undefined
+      ],
+      [
+        'pip',
+        '3.8.12',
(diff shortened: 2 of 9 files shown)


oranav:
I still need to test this locally, hence this is a draft.

oranav:
@IvanZosimov Sorry for the long delay! This is finally ready for review. All tests are passing successfully, and I'm using my own branch as an action in another project.

IvanZosimov:
> I fixed the failing main workflow on Windows. Good catch.
> 
> That being said, I don't understand why e2e-cache fails with Pipenv and Windows. I haven't changed anything related to it. Can it be failing randomly?

Hi, @oranav 👋 Thank you for this PR and sorry for the delay! The problem with e2e test related to pipenv on Windows is most probably caused by recent pipenv minor tag update, I hope that they will repair it. You also have 3 failing unit-tests in Main workflow, could take a look at them?

oranav:
It seems that I've hit an unrelated bug that happens sporadically, which causes tests to timeout in some cases. Take a look at [this action for example](https://github.com/actions/setup-python/actions/runs/3244002148/jobs/5319547229), which has failed for the same reason.

I've fixed it in [this commit](https://github.com/oranav/setup-python/commit/9578b65db127c898fb1dd4de1d68475e473de542). In short, the cache distributor tests were looking for dependency file with their default glob (`**/requirements.txt` for example), which made it search for the file in the **project's root** instead of in the tests data folder. I've changed that, and it reduced the tests time on my machine from ~7s to less than 3 seconds.

After we run the workflows again I expect the main workflow to pass.

oranav:
Hey @IvanZosimov, can you please take a look?

oranav:
I'm not sure why main workflow failed with Python 2.7 (which I haven't changed at all); I've seen that it failed in [other](https://github.com/actions/setup-python/actions/runs/3359996628/jobs/5570183786#step:7:10) [runs](https://github.com/actions/setup-python/actions/runs/3352973492/jobs/5555472793#step:7:16) as well, so I suspect it's unrelated to my PR.
I rebased my commits on top of `main`. Can we please re-run?

oranav:
@dsame @IvanZosimov Can we please re-run the actions?

oranav:
Sorry again. The problem was indeed unrelated to my PR, and it seems that #535 fixed the problem (for the curious - there is no Python 2.7 for Ubuntu 22.04, and the workflow previously used `ubuntu-latest` which got upgraded to 22.04). Unfortunately, #535 wasn't included in my rebase; I now rebased on top of it as well. Running again should result in a working workflow. Can we please re-run one more time?

I ran the workflow in my repo just in case, and it passed successfully:
https://github.com/oranav/setup-python/actions/runs/3412737221/jobs/5678633730

oranav:
@dsame @IvanZosimov Can we please re-run the workflow?
Sorry for the back and forth :)

dsame:
@oranav can you please explain the purpose of running `poetry env use` after the `.venv` already has been added to the list of cached directories? Is it just to create `.venv` directory? 

i mean this https://github.com/actions/setup-python/pull/447/files#diff-5a0a81991280dd0bbc75f051187afdb8ca7cb621a0bbc95b0d8d5d9a30f143a1R48

oranav:
@dsame It's actually the same code as before, only in a loop (that iterates over all nested projects).
I believe `poetry env use` tells Poetry which Python binary to use; since there might not be any cached venv (if it's the first run, or if it has been evicted), then it'll initialize the venv with the correct Python binary.

__tests__/cache-restore.test.ts
```diff
@@ -99,21 +101,68 @@ virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/patrick/Library/Caches/py
...
       [
         'pip',
         '3.8.12',
         '__tests__/data/requirements-linux.txt',
-        requirementsLinuxHash
+        requirementsLinuxHash,
+        undefined
       ],
-      ['pip', '3.8.12', '__tests__/data/requirements.txt', requirementsHash],
-      ['pipenv', '3.9.1', undefined, pipFileLockHash],
```
MaksimZhukov, line 111: Could you please clarify why you have decided to remove the test cases where the `dependencyFile` parameter is set to `undefined`? They are intended to test the default behavior of action.

MaksimZhukov:
Hello @oranav!
Thank you for the contribution! LGTM!

I think [documentation](https://github.com/actions/setup-python#caching-packages-dependencies) needs to be updated as well. At the moment it says that only `virtualenv` directory will be cached

__tests__/cache-restore.test.ts
MaksimZhukov: Sorry, I haven't noticed [your comment](https://github.com/actions/setup-python/pull/447#issuecomment-1278952666)

src/cache-distributions/poetry-cache.ts
```diff
@@ -16,41 +16,49 @@ class PoetryCache extends CacheDistributor {
...
+      const virtualenvsPath = poetryConfig['virtualenvs.path'].replace(
+        '{cache-dir}',
+        cacheDir
       );
 
-      if (exitCode) {
-        logWarning(stderr);
+      paths.add(virtualenvsPath);
+
+      if (poetryConfig['virtualenvs.in-project'] === true) {
```
dmitry-shibanov, line 44: if it is always a bool value I think we can change it to `if (poetryConfig['virtualenvs.in-project'])`

src/cache-distributions/poetry-cache.ts
```diff
@@ -16,41 +16,49 @@ class PoetryCache extends CacheDistributor {
...
+        if (exitCode) {
+          logWarning(stderr);
+        }
       }
-    } else {
-      logWarning('python binaries were not found in PATH');
     }
 
-    return paths;
+    return Array.from(paths);
```
dmitry-shibanov, line 61: Could you please change it to `[...paths]`.

README.md
```diff
@@ -55,7 +55,7 @@ The action defaults to searching for a dependency file (`requirements.txt` for p
 
  - For `pip`, the action will cache the global cache directory
  - For `pipenv`, the action will cache virtualenv directory
- - For `poetry`, the action will cache virtualenv directory
+ - For `poetry`, the action will cache virtualenv directories -- one for each poetry project found
```
MaksimZhukov, line 58: I think this is a clearer description:

```suggestion
 - For `poetry`, the action will cache virtualenv directory. If the `virtualenvs.in-project` setting is set to `true`, the action will also cache the `.venv` folder within the root directory of each poetry project found. 
```

__tests__/cache-restore.test.ts
MaksimZhukov: I think we need to revert these cases to keep on testing the default behavior.

README.md
oranav: That's not the only case though. If `virtualenvs.path` is customized, then it will also cache another folder.
How about the following?

```suggestion
 - For `poetry`, the action will cache virtualenv directories. Since poetry projects may customize the virtualenv path (using `virtualenvs.in-project` and `virtualenvs.path`), this may result in multiple folders being cached when there are multiple poetry projects.
```

MaksimZhukov: Sounds good to me

__tests__/cache-restore.test.ts
oranav: Again, we're testing the exact same test cases as before. The only difference is that we're looking for lock files inside `__tests__/data/` (instead of looking also inside the source code folder, which obviously doesn't make sense).

If we revert this behavior then tests will continue to fail sporadically, as happened before.

MaksimZhukov (approved):

marko-zivic-93 (approved):

dmitry-shibanov (approved):

brcrista (commented):
Thanks for the PR @oranav.  Did you test the compat case -- saving a cache with the existing version of `setup-python` and restoring with this change?

dsame:
> Thanks for the PR @oranav. Did you test the compat case -- saving a cache with the existing version of `setup-python` and restoring with this change?

@oranav please response this comment, we need it to merge the PR

oranav:
@brcrista @dsame Sorry, I'm out of home until Sunday. I did **not** test the compat case -- good catch! It _might_ be broken since we're saving multiple directories instead of one, but I don't believe it is.

I can test this as soon as I get back on Sunday. Is there an automated test of some kind?

brcrista:
@oranav no, you'll have to test it manually.

agucova:
@oranav heads up that the compat case doesn't seem to work. In my testing, this update triggered Poetry to not be able to find the correct python version (even after the `poetry env use` being run). Clearing the cache fixed it.

oranav:
@brcrista @dsame @agucova I've tested the compat case thoroughly. Here's a summary of the results:

| Poetry located at | `virtualenvs.in-project` | Pre-PR workflow | Post-PR workflow |
| ----------------- | ------------------------ | ------------------| ------------------- |
| Root | `false` | [Pass](https://github.com/oranav/setup-python/actions/runs/3609466796) | [Pass](https://github.com/oranav/setup-python/actions/runs/3609472754) |
| Root | `true` | [Pass](https://github.com/oranav/setup-python/actions/runs/3609494774) | [Pass](https://github.com/oranav/setup-python/actions/runs/3609496831) |
| Inner folder | `false` | [Failed silently, cache saved](https://github.com/oranav/setup-python/actions/runs/3609504119) (wrong Python version used due to wrong `poetry env use`) | [Fail with old cache](https://github.com/oranav/setup-python/actions/runs/3609505528) (Failed because it loaded an old, corrupt cache; will pass after invalidating the cache) |
| Inner folder | `true` | [Fail](https://github.com/oranav/setup-python/actions/runs/3609522573) | [Pass](https://github.com/oranav/setup-python/actions/runs/3609526878) |

As you can see, the only interesting case is number 3, in which if the previous version of the workflow was run, then a corrupt cache was already saved, and the post-PR version will happily load this cache and will fail with a wrong Python version (because the virtualenv was built incorrectly).

I see two options:

1. Do nothing. This actually makes sense, since the action did not work properly even before; it seems as though the action "passed", but it actually failed because it has built a virtualenv with the wrong Python version.
2. Change the [Poetry cache primary key](https://github.com/actions/setup-python/blob/main/src/cache-distributions/poetry-cache.ts#L58) such that **all** previous caches are invalidated, which will make everything pass happily.

I'm leaning towards option 1, since option 2 solves a very esoteric case. Thoughts?

agucova:
I was on the “fail silently” option and I think it makes sense to try to repair it a posteriori, especially because that happened when the correct setup was being used.

It's difficult to estimate, but I wouldn't be surprised if it was the case that there were lots of people using the action without the cache working.

brcrista:
@oranav thank you for testing that!  The two things we want to avoid are, in priority order:
1. Workflows that were passing to failing after the change
2. Workflows that were getting cache hits to start getting cache misses after the change (but still succeeding)

It looks like the interesting case there is

| Poetry located at | `virtualenvs.in-project` | Pre-PR workflow | Post-PR workflow |
| ----------------- | ------------------------ | ------------------| ------------------- |
| Inner folder | `false` | [Failed silently, cache saved](https://github.com/oranav/setup-python/actions/runs/3609504119) (wrong Python version used due to wrong `poetry env use`) | [Fail with old cache](https://github.com/oranav/setup-python/actions/runs/3609505528) (Failed because it loaded an old, corrupt cache; will pass after invalidating the cache) |

I see in the [logs](https://github.com/oranav/setup-python/actions/runs/3609504119/jobs/6082735764#step:5:21) for the Pre-PR workflow:

```
The currently activated Python version 3.8.10 is not supported by the project (^3.10).
Trying to find and use a compatible version. 
Using python3 (3.10.8)
```

so before the run is passing but caching is broken, and after the run is failing.  Caching should be transparent: if the cache is corrupt, we should ignore the cache, invalidate it, and create a new one.

How much work would it be to handle that case?  Since this should only happen on upgrade, we can handle it with a new major version release of the action.  But that would slow down adoption of this feature, so it's better to handle it if we can.

oranav:
Woah, I finally triaged the issue. The old action never ran `poetry env use` correctly, so the `envs.toml` file was never created; this in turn made Poetry [ignore the venv](https://github.com/python-poetry/poetry/blob/f594246c9fd70ce9c9379afb45c63c19f9123fa0/src/poetry/utils/env.py#L699) in case it is broken.
With my fix applied, `poetry env use` is run correctly, therefore `envs.toml` is created, but the virtualenv directory already exists (from the cache), so it [forces Poetry to use the broken venv](https://github.com/python-poetry/poetry/blob/f594246c9fd70ce9c9379afb45c63c19f9123fa0/src/poetry/utils/env.py#L896-L904).

Either way, we'll want to cache to be rewritten after this PR. I suggest that we'll just bump the Poetry cache distributor primary key, so users will receive a cache miss upon the first run of the new action, and from there everything will work as expected.

I've pushed my changes. Can we please re-run the workflows?

oranav:
Just to be clear: I've re-run the failing case ([before](https://github.com/oranav/setup-python/actions/runs/3771798616/jobs/6412379629), [after](https://github.com/oranav/setup-python/actions/runs/3771820961/jobs/6412420186)) and it looks okay now -- the "after" case is working correctly, because it doesn't even use the old cache.

oranav:
Anything still missing from my side or can we proceed to merging this? 😊

oranav:
@brcrista @dsame @agucova Forgot to tag you in the previous comments, trying again in case there were no notifications. :)

brcrista (approved):
Looks good @oranav !  I think that's a good fix for the compat issue.  Thanks for the contribution!

invisiblepancake (commented):
src/setup-python.ts
The file sets up the Python environment with caching capabilities.
No code appears suspicious or malicious; it primarily uses core modules and standard libraries.
package.json
The file defines dependencies, scripts, and metadata for the project.
Dependencies and scripts appear standard and necessary for the project.
src/utils.ts
This file includes utility functions for handling system-specific tasks and managing versions.
No code appears suspicious or malicious; it includes standard utility functions.

src/cache.ts

Failed to retrieve the file changes.

invisiblepancake (commented):
Changes Made:

Added a basedir parameter to set the current working directory for the poetry command.

Potential Issues:
No apparent malicious code. The changes ensure the command runs in the correct directory context.

Summary:

The changes introduced in the functions are aimed at improving the handling of multiple Poetry projects by dynamically identifying and processing their configurations.

No suspicious or malicious code patterns were identified.

invisiblepancake (commented):
chwers

invisiblepancake (commented):
README.md


Changes Made:

Updated documentation to explain that for poetry, the action will cache virtualenv directories for each poetry project found.
Potential Issues: No issues identified; the change clarifies the caching behavior for poetry.

tests/cache-restore.test.ts



Changes Made:

Added imports for path and State.
Corrected a typo from saveSatetSpy to saveStateSpy.
Updated test cases to include new paths and hash values.
Enhanced test cases to verify that saveState is called with the correct cache paths.

Potential Issues: No issues identified; the changes improve test coverage and fix typos.

invisiblepancake (commented):
More tuning. hf =)

invisiblepancake (commented):
When collecting Poetry projects for caching, a '**/poetry.lock' glob is used.  However, in order to process the Poetry configuration, the"poetry" command is run from the repo's root directory; this causes
Poetry to return an invalid configuration when there is a Poetry project
inside an inner directory.

Instead of running a single Poetry command, glob for the same pattern,
and run a Poetry command for every discovered project.

invisiblepancake (commented):
Simplified the condition check by removing the explicit comparison to true.

Issue #432, opened 2022-06-12
Dependencies cache key has to be specific to OS version


illia-v:
**Description:**

A cache key looks like this `setup-python-Linux-python-3.11.0-beta.3-pip-60d374041db12689c3d346e149aa43744f06aa781ab1c8cbbe70e0b09b98dd70`.
It is specific to a Python version and OS type. It has to be specific to a version of an operating system too.

**Justification:**

PyPI provides source distributions of packages (archives with code) and built distributions (wheels). When a wheel suitable for an environment is not found on PyPI, a source distribution is usually downloaded and a wheel is built locally. The source may contain some code that needs to be compiled (C, Rust, etc.)

On Linux, a compiled wheel may depend on a specific version of a shared library. Different Ubuntu releases used on GitHub Actions may provide different versions of shared libraries. Therefore, a wheel compiled on Ubuntu 20.04 may be incompatible with Ubuntu 22.04. This makes the cache feature unusable when a workflow runs jobs on multiple Ubuntu versions or when an Ubuntu version is changed in a workflow.

https://github.com/urllib3/urllib3/runs/6817232938 is an example of a failed run.
The error is caused by a [cffi](https://pypi.org/project/cffi/) wheel that was compiled and cached in Ubuntu 20.04 previously where it depended on `libffi.so.7`.
Ubuntu 22.04 (that provides `libffi.so.8`) restored that cache archive because it composed the same cache key, and cffi become unusable there (https://github.com/urllib3/urllib3/pull/2626#issuecomment-1153002444 provides a little bit more info.)

BTW, #188 may be a related bug report issue.

**Are you willing to submit a PR?**

No. Although, I guess it is a matter of changing the following lines of code to include a version of an OS in the key.
https://github.com/actions/setup-python/blob/813f9b1556d9f0a0faf7b08a19486d6af9857537/src/cache-distributions/pip-cache.ts#L60-L61

dsame:
@e-korolevskii implementing this feature should solve [this](https://github.com/actions/setup-python/issues/188) issue and [this](https://github.com/actions/setup-python/issues/442)

illia-v:
@panticmilos thank you!

illia-v:
@panticmilos what do you think about updating these values too?
https://github.com/actions/setup-python/blob/592a7a7a45a48972db5452787b01b1f407e38948/src/cache-distributions/poetry-cache.ts#L36
[...]

PR #462, opened 2022-07-15
Add Linux version and release to primary key

panticmilos:
**Description:**
This PR provides expanded cache primary and restore keys for Linux distributions

**Related issue:**
https://github.com/actions/setup-python/issues/432

**Check list:**
- [x ] Mark if documentation changes are required.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -3,6 +3,7 @@ import * as core from '@actions/core';
 import fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
+import * as exec from '@actions/exec';
 
 export const IS_WINDOWS = process.platform === 'win32';
 export const IS_LINUX = process.platform === 'linux';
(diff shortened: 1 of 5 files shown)


vsafonkin (approved):

marko-zivic-93 (approved):

src/utils.ts
```diff
@@ -119,3 +120,18 @@ export function isCacheFeatureAvailable(): boolean {
 
   return true;
 }
+
+export async function getLinuxOSReleaseInfo() {
+  const versionId = await exec.getExecOutput('lsb_release', ['-a']);
```
IvanZosimov, line 125: I'd suggest to add option [silent](https://github.com/actions/toolkit/blob/main/packages/exec/src/interfaces.ts#L13) here because I think we don't want a user to see the output of `lsb_release -a` in logs. See the screenshot of action outputs as it is now:  
image

Issue #184, opened 2021-01-25
Tracking: On GHES, there is no fallback if this action is rate limited


yacaovsnc:
**Describe the bug**
On GHES, we do not use the token input from user.  So if this action is rate limited, there is no workaround. 

**Which version of the action are you using?**

- [ ] `v1`
- [x] `v2`
- [ ] Some other tag (such as `v2.0.1` or `master`)

killuazhu:
We run into the same issue... @yacaovsnc did you just ask your users to put a GH token in the action?

dsame:
@yacaovsnc  can you please describe your situation:

What is the reason it is not possible to use token input?
Which workaround do you prefer to have?

scream314:
@dsame I guess the desired behaviour would be something like in case of https://github.com/actions/setup-node.

`setup-node`:
```
Run actions/setup-node@v3
Attempting to download 14...
API rate limit exceeded for 1.2.3.4. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
Falling back to download directly from Node
Acquiring 14.20.0 - x64 from https://nodejs.org/dist/v14.20.0/node-v14.20.0-linux-x64.tar.gz
Extracting ...
```
(and then it continues after extracting)

`setup-python`:
```
Run actions/setup-python@v3
Version 3.x was not found in the local cache
Error: API rate limit exceeded for 1.2.3.4. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```
(the action just stops after the rate limit error)

> What is the reason it is not possible to use token input?

I'm not even sure what token should be used and where, many of the users do not have a github.com account, and in case of automated workflows there is no user involved at all.

> Which workaround do you prefer to have?

Falling back to eg. python.com (the same way as `setup-node` falls back to nodejs.org.

aboutthatjazz:
A friendly bump as this issue if forcing customers to workaround this issue in a manner that increases the complexity to provide support for our customers.

zetaab:
latest GHES https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-setup-python-on-ghes instruction says that we should define github.com token in case of rate limits. However, that does not work either.

workflow:

```
    - name: Caching Python 3.8
      uses: actions/setup-python@v4
      with:
        token: ${{ secrets.GH_DOTCOM_TOKEN }}
        python-version: 3.8
```

logs:

```
[...]

PR #511, opened 2022-09-28
Handle download HTTP error

dsame:
**Description:**

Handle the error during an attempt to download python archive in the manner like `setup-node` [does](https://github.com/actions/setup-node/blob/969bd2663942d722d85b6a8626225850c2f7be4b/src/installer.ts#L109).

```
Run actions/setup-node@v3
Attempting to download 14...
API rate limit exceeded for 1.2.3.4. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
Falling back to download directly from Node
Acquiring 14.20.0 - x64 from https://nodejs.org/dist/v14.20.0/node-v14.20.0-linux-x64.tar.gz
Extracting ...
```

currently setup-node is just stops
```
Run actions/setup-python@v3
Version 3.x was not found in the local cache
Error: API rate limit exceeded for 1.2.3.4. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

**Related issue:**
[Add link to the related issue.](https://github.com/actions/setup-python/issues/184)

**Check list:**

### DIFF
--- a/src/install-pypy.ts
+++ b/src/install-pypy.ts
@@ -46,37 +46,58 @@ export async function installPyPy(
   let downloadUrl = `${foundAsset.download_url}`;
 
   core.info(`Downloading PyPy from "${downloadUrl}" ...`);
-  const pypyPath = await tc.downloadTool(downloadUrl);
 
-  core.info('Extracting downloaded archive...');
-  if (IS_WINDOWS) {
-    downloadDir = await tc.extractZip(pypyPath);
-  } else {
-    downloadDir = await tc.extractTar(pypyPath, undefined, 'x');
+  try {
+    const pypyPath = await tc.downloadTool(downloadUrl);
+
+    core.info('Extracting downloaded archive...');
+    if (IS_WINDOWS) {
+      downloadDir = await tc.extractZip(pypyPath);
+    } else {
+      downloadDir = await tc.extractTar(pypyPath, undefined, 'x');
+    }
+
+    // root folder in archive can have unpredictable name so just take the first folder
+    // downloadDir is unique folder under TEMP and can't contain any other folders
+    const archiveName = fs.readdirSync(downloadDir)[0];
+
+    const toolDir = path.join(downloadDir, archiveName);
+    let installDir = toolDir;
+    if (!isNightlyKeyword(resolvedPyPyVersion)) {
+      installDir = await tc.cacheDir(
+        toolDir,
+        'PyPy',
+        resolvedPythonVersion,
+        architecture
+      );
+    }
+
+    writeExactPyPyVersionFile(installDir, resolvedPyPyVersion);
+
+    const binaryPath = getPyPyBinaryPath(installDir);
+    await createPyPySymlink(binaryPath, resolvedPythonVersion);
+    await installPip(binaryPath);
+
+    return {installDir, resolvedPythonVersion, resolvedPyPyVersion};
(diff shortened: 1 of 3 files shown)


marko-zivic-93 (approved):

src/install-pypy.ts
```diff
@@ -47,37 +47,58 @@ export async function installPyPy(
...
+    return {installDir, resolvedPythonVersion, resolvedPyPyVersion};
+  } catch (err) {
+    if (err instanceof Error) {
+      // Rate limit?
+      if (
+        err instanceof tc.HTTPError &&
+        (err.httpStatusCode === 403 || err.httpStatusCode === 429)
+      ) {
+        core.info(
+          `Received HTTP status code ${err.httpStatusCode}.  This usually indicates the rate limit has been exceeded`
```
brcrista, line 91: This makes sense for 429.  But I would expect 403 to be an expired PAT or insufficient permissions?

brcrista (approved):
Discussed offline -- `setup-node` has the same message for the same condition.  We'll check if `403` is correct here and update both actions if needed.

src/install-pypy.ts
dsame: As discussed we should keep the code coherent with others actions and improve the handling 423/403 error codes within another issue.

vsafonkin (approved):

brcrista:
> We'll check if `403` is correct here

I had sent this to @dsame on Slack, but I'll repeat it here.  When `GITHUB_TOKEN` gets rate limited, it will show up as a 403.  Apparently 429 wasn't a response code yet when GitHub added rate limiting.

src/install-python.ts
```diff
@@ -72,15 +72,36 @@ export async function installCpythonFromRelease(release: tc.IToolRelease) {
...
+    } else {
+      pythonExtractedFolder = await tc.extractTar(pythonPath);
+    }
 
-  core.info('Execute installation script');
-  await installPython(pythonExtractedFolder);
+    core.info('Execute installation script');
+    await installPython(pythonExtractedFolder);
+  } catch (err) {
+    if (err instanceof Error) {
```
dmitry-shibanov, line 89: Do we need this check ? I think we can remove it.

src/install-python.ts
```diff
@@ -72,15 +72,36 @@ export async function installCpythonFromRelease(release: tc.IToolRelease) {
...
+        err instanceof tc.HTTPError &&
+        (err.httpStatusCode === 403 || err.httpStatusCode === 429)
+      ) {
+        core.info(
+          `Received HTTP status code ${err.httpStatusCode}.  This usually indicates the rate limit has been exceeded`
+        );
+      } else {
+        core.info(err.message);
+      }
+      if (err.stack !== undefined) {
```
dmitry-shibanov, line 101: could you please change it to `if (err.stack) {`

randydwhitephotography (approved):

randydwhitephotography (approved):

randydwhitephotography (approved):

dmitry-shibanov (approved):

randydwhitephotography (approved):

randydwhitephotography (approved):

randydwhitephotography (approved):

marko-zivic-93 (approved):

PR #545, opened 2022-11-21
Recommend setting python-version

heavywatal:
**Description:**
An expression in `README.md` was fixed. It does not make sense to recommend using `setup-python` in this context.

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ steps:
     python-version: 'pypy3.9' 
 - run: python my_script.py
 ```
-The `python-version` input is optional. If not supplied, the action will try to resolve the version from the default `.python-version` file. If the `.python-version` file doesn't exist Python or PyPy version from the PATH will be used. The default version of Python or PyPy in PATH varies between runners and can be changed unexpectedly so we recommend always using `setup-python`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,7 +33,7 @@ steps:
     python-version: 'pypy3.9' 
 - run: python my_script.py
 ```
-The `python-version` input is optional. If not supplied, the action will try to resolve the version from the default `.python-version` file. If the `.python-version` file doesn't exist Python or PyPy version from the PATH will be used. The default version of Python or PyPy in PATH varies between runners and can be changed unexpectedly so we recommend always using `setup-python`.
+The `python-version` input is optional. If not supplied, the action will try to resolve the version from the default `.python-version` file. If the `.python-version` file doesn't exist Python or PyPy version from the PATH will be used. The default version of Python or PyPy in PATH varies between runners and can be changed unexpectedly so we recommend always setting `python-version` explicitly.
```
MaksimZhukov, line 36: Hello @heavywatal! Thank you for the contribution!
I think we need to recommend both options.
```suggestion
The `python-version` input is optional. If not supplied, the action will try to resolve the version from the default `.python-version` file. If the `.python-version` file doesn't exist Python or PyPy version from the PATH will be used. The default version of Python or PyPy in PATH varies between runners and can be changed unexpectedly so we recommend always setting Python version explicitly using the `python-version` or `python-version-file` inputs.
```

heavywatal: Committed accordingly. Thank you for the suggestion.

MaksimZhukov (approved):

dmitry-shibanov (approved):

PR #549, opened 2022-11-28
Caching projects that use setup.py

simonw:
My Python projects generally use a `setup.py` file to manage their dependencies - it wasn't completely obvious from this document how to run those, so I added an extra example.

See also my TIL: https://til.simonwillison.net/github-actions/cache-setup-py


brcrista (approved):
Thanks @simonw!

Issue #547, opened 2022-11-22
Provide x64, arm64, and universal2 arch selectors for darwin


mattip:
**Description:**
Describe your proposal.

As described in [this comment](https://github.com/actions/python-versions/pull/114#issuecomment-1193097340) by @misl6, it would be nice to add the possibility to choose between `arch = universal2, arm64, x64` on darwin (macos).

**Justification:**
Justification or a use case for your proposal.

Runners for arm64 macos are [in the github roadmap](https://github.com/github/roadmap/issues/528), but are not generally available. Once they are, it would be nice to be able to use either "thin" (x86_64 or arm64) or universal2 pythons on that image. Some work toward this was done in PR #114 which was merged then reverted. The design for the arch selector should already be worked out before making the images available:
- what is the exact syntax of the selector
- what will the default be

Note that PyPy is going to release both thin arm64  and x86_64 binary packages in its next release, and CPython already has a universal2 package. 

I would propose supporting all three alternatives, as the scientific python stack (NumPy, SciPy, and more) and conda-forge have decided to support thin binary packaging only. It would be convenient if this action would make it easy to support those decisions by not forcing a universal2-only solution.

xref issue #108 which discusses the need for arm/arm64 python.

**Are you willing to submit a PR?**

Yes, but I don't know if it is relevant for an outside contribution.

MaksimZhukov:
Hello @mattip! Thank you for the suggested idea!
We will consider adding this feature and will let you know as soon as we have any decision.

mattip:
The link you posted states that "__self-hosted__ runners are available". That is great, and is probably enough to justify formalizing the syntax for the selectors now. I was relating to the possibility of __github-hosted__ runners, which seems to be not generally available yet.

jdogmcsteezy:
@mattip, I should have thought about it a little harder, haha.  Thanks for the clarification

NeilJed:
+1 To this. I'm building an Ubuntu 22.04 runner on ARM64 from the runner-images repo and the lack of ARM64 toolcache packages makes it impossible to use this action. TBH all the "setup/xxx" toolcache packages should be multi-arch.

mattip:
@mayeut: nice. 
FWIW, the run you pointed to above [seems to use x86_64](https://github.com/mayeut/setup-python/actions/runs/3669884297/jobs/6204031022#step:4:11) for the CPython run. Is that intentional?

mayeut:
> Is that intentional?

[...]

PR #394, opened 2022-04-28
Assign default value of AGENT_TOOLSDIRECTORY if not set

dsame:
**Description:**
This is a proof of concept PR to eliminate the need of explicit setting  AGENT_TOOLSDIRECTORY variable

Ubuntu demo build 
https://github.com/akv-demo/setup-python-test/runs/6205091860?check_suite_focus=true

**Related issue:**
Add link to the related issue.

**Check list:**
- [ ] Mark if documentation changes are required.
- [ ] Mark if tests were added or updated to cover the changes.

PR #564, opened 2022-12-11
fix: don't override `AGENT_TOOLSDIRECTORY` if it's not needed

mayeut:
**Description:**
Don't override `AGENT_TOOLSDIRECTORY` if it's not needed:

`AGENT_TOOLSDIRECTORY` is always overriden on macOS.
This seems to be needed because CPython < 3.11 x64 builds are not relocatable
 It means this is not needed for PyPy and also not needed when targetting macOS arm64 runners.

**Related issue:**
Towards #547 

**Check list:**

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -66896,21 +66896,23 @@ function resolveVersionInput() {
 function run() {
     var _a;
     return __awaiter(this, void 0, void 0, function* () {
-        if (utils_1.IS_MAC) {
-            process.env['AGENT_TOOLSDIRECTORY'] = '/Users/runner/hostedtoolcache';
-        }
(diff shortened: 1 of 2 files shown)


brcrista:
@dsame could you have a look at this?  It's related to https://github.com/actions/setup-python/pull/394

I believe [this section of the docs](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#macos) explains why we do this.

@mayeut I understand why we don't need it for PyPy, but could you explain why it's not needed for macOS arm64 runners?

mayeut:
This would need a full rework with the new multi versions support.
Closing for now.

> but could you explain why it's not needed for macOS arm64 runners?

@brcrista, if using universal2 installers as proposed in https://github.com/actions/python-versions/pull/214, then the python installation is always installed in a fixed framework location and the tool cache only manages symlinks. We can create those symlinks wherever we want.

PR #595, opened 2023-01-31
Remove .python-version warnings

stevelacey:
I suggest removing these warnings because they imply something is wrong when there probably isn't, and other `setup-{lang}` actions don't have them; the relevant information is already shown on the next line when using the functionality.

Screen Shot 2023-01-31 at 19 30 55

The only scenario that really matters here is the happy path, which is already covered by https://github.com/actions/setup-python/blob/1811840/src/setup-python.ts#L46 informing the dev what happened when something did happen.

The unhappy path, where the file does not exist is already covered in https://github.com/actions/setup-python/blob/1811840/src/setup-python.ts#L118-L120.

The scenarios where `python-version` or `python-version-file` are supplied don't change and will continue to warn/error if they have bad values.

Note that the second warning I've removed only fires when there is no file input provided aka the value will always be the default of `.python-version`.

No documentation or tests cover these warnings, so there's nothing to update.

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -66942,14 +66942,12 @@ function resolveVersionInput() {
         core.info(`Resolved ${versionFile} as ${version}`);
         return [version];
     }
-    utils_1.logWarning("Neither 'python-version' nor 'python-version-file' inputs were supplied. Attempting to find '.python-version' file.");
     versionFile = '.python-version';
     if (fs_1.default.existsSync(versionFile)) {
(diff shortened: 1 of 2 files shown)


sanjeebtiwary (approved):

dmitry-shibanov:
Hello @stevelacey. Sorry for the late response. I think it will be better to keep this warnings because they were added in scope of this [pull request ](https://github.com/actions/setup-python/pull/450) to improve messages. Besides, I'd like to keep them because they indicate about some breaking changes in v4 and it helps to transit from older versions.

PR #611, opened 2023-02-13
Output `python-versions`, containing all Python versions, for cache-busting

### LINKED ISSUE #606: Identify all installed Python versions
### **Description:**

I'd like trivially use all of the installed Python versions (in full) in my cache keys (like "3.8.16" and "3.9.16 (PyPy 7.3.11)").

This could be implemented in several ways, including:

* A `python-versions` (plural!) output that includes the full Python versions in full. (How it's delimited doesn't matter to me due to my use case.) For example, `3.8.10; 3.9.13; 3.9.16 (PyPy 7.3.11)`.
* A file written at a known or configurable location that lists all of the install Python versions.

### **Justification:**

I'm very pleased with the new ability to install multiple Python versions simultaneously! This pairs well with tox-based testing.

I'm wanting to cache and restore my `.tox/` directories between runs, particularly on Windows hosts, because it's expensive to set up the tox environments for each Python version. In the past, with only one Python version installed, I would use the action output, `python-version`, as a part of the cache key.

However, when multiple versions are installed, if there is a cache hit but one of the Python versions that isn't the default has updated (say, if "3.11" is the default but "3.8" updates from "3.8.10" to "3.8.11") then tox will rebuild the 3.8 environment each time the cache is restored, the cache won't be invalidated, and test setup times will start to increase.

I don't know of a trivial, cross-platform way to write this myself that will work consistently as new Python versions and host platforms are added (like Pyston, as I saw in another ticket)...but I'd prefer to avoid crawling the filesystem and executing `python -V` on every executable that looks like a Python interpreter.

Therefore, I think that outputting this metadata from within the setup-python action would be the most ideal scenario.

### **Are you willing to submit a PR?**

I'd love to try! I don't know TypeScript and it might take a bit to write something that fits the requirements _and_ runs correctly.

I'll take a look at what that might look like (particularly by studying the existing `python-version` output) while awaiting feedback for this idea.

kurtmckee:
**Description:**
This introduces a `python-versions` output, which is a comma-separated string of all installed Python versions and implementations. This is useful for cache-busting when multiple Python versions are installed, as described in #606.

For example, it is now possible to cache a `.tox/` test directory and invalidate the cache when _any_ of the Python versions change. Caching the `.tox/` directory offers a significant speedup to test setup times, but if a non-default Python's version changes (say, if 3.11.1 is the default but 3.10.x increments to 3.10.y), it is currently very difficult to invalidate the cache. This new output variable allows for trivial cache invalidation in this situation.

To demonstrate the new output variable, I created a workflow with a single `setup-python` step that installs multiple CPython and PyPy versions, then used `echo` to [show the value of `python-versions`](https://github.com/kurtmckee/test-setup-python-python-versions/actions/runs/4164961852/jobs/7207451981#step:3:11).

The value displayed is:

```
3.10.9,3.11.1,3.12.0-alpha.5,3.7.15,3.8.16,3.8.16-pypy7.3.11,3.9.16,3.9.16-pypy7.3.11
```

which can be used directly in a cache key, or written to a file and included in a call to `hashFiles()`. The versions are sorted to avoid sensitivity to the order of Python versions specific in the workflow, but are sorted merely as strings for this purpose, not as semantic versions. Therefore a user can create a workflow that installs Python versions in any order they want, but the `python-versions` output string will be stable.

NOTE: This work depends on #610 to remove trailing newlines in PyPy versions.

**Related issue:**
Closes #606 

**Check list:**

### DIFF
--- a/src/setup-python.ts
+++ b/src/setup-python.ts
@@ -95,6 +95,7 @@ async function run() {
 
     if (versions.length) {
       let pythonVersion = '';
+      const pythonVersions: string[] = [];
       const arch: string = core.getInput('architecture') || os.arch();
       const updateEnvironment = core.getBooleanInput('update-environment');
       core.startGroup('Installed versions');
@@ -108,6 +109,9 @@ async function run() {
             allowPreReleases
           );
           pythonVersion = `${installed.resolvedPyPyVersion}-${installed.resolvedPythonVersion}`;
+          pythonVersions.push(
+            `${installed.resolvedPythonVersion}-pypy${installed.resolvedPyPyVersion}`
+          );
           core.info(
             `Successfully set up PyPy ${installed.resolvedPyPyVersion} with Python (${installed.resolvedPythonVersion})`
           );
@@ -120,6 +124,7 @@ async function run() {
             allowPreReleases
           );
           pythonVersion = `${installed}`;
+          pythonVersions.push(`graalpy${installed}`);
           core.info(`Successfully set up GraalPy ${installed}`);
         } else {
           if (version.startsWith('2')) {
@@ -135,9 +140,11 @@ async function run() {
             allowPreReleases
           );
           pythonVersion = installed.version;
+          pythonVersions.push(installed.version);
           core.info(`Successfully set up ${installed.impl} (${pythonVersion})`);
         }
       }
+      core.setOutput('python-versions', pythonVersions.sort().join(','));
       core.endGroup();
       const cache = core.getInput('cache');
       if (cache && isCacheFeatureAvailable()) {
--- a/action.yml
+++ b/action.yml
@@ -29,6 +29,8 @@ inputs:
 outputs:
   python-version:
     description: "The installed Python or PyPy version. Useful when given a version range as input."
+  python-versions:
+    description: "A comma-separated list of all installed Python implementations and versions. Useful for cache-busting."
   cache-hit:
     description: "A boolean value to indicate a cache entry was found"
   python-path:
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -70208,6 +70208,7 @@ function run() {
             const allowPreReleases = core.getBooleanInput('allow-prereleases');
             if (versions.length) {
                 let pythonVersion = '';
+                const pythonVersions = [];
                 const arch = core.getInput('architecture') || os.arch();
                 const updateEnvironment = core.getBooleanInput('update-environment');
                 core.startGroup('Installed versions');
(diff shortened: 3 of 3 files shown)


src/setup-python.ts
```diff
@@ -94,6 +95,9 @@ async function run() {
             allowPreReleases
           );
           pythonVersion = `${installed.resolvedPyPyVersion}-${installed.resolvedPythonVersion}`;
+          pythonVersions.push(
+            `${installed.resolvedPythonVersion}-pypy${installed.resolvedPyPyVersion}`
+          );
```
dmitry-shibanov, line 100: Do we need to change pypy and python version parts order ?

src/setup-python.ts
```diff
@@ -106,9 +110,11 @@ async function run() {
             allowPreReleases
           );
           pythonVersion = installed.version;
+          pythonVersions.push(installed.version);
           core.info(`Successfully set up ${installed.impl} (${pythonVersion})`);
         }
       }
+      core.setOutput('python-versions', pythonVersions.sort().join(','));
```
dmitry-shibanov, line 117: Do we need to sort versions ?

src/setup-python.ts
kurtmckee: It's not a need, no!

I did it this way for potential parsing purposes. It seemed like if someone wanted to determine the Python version (regardless of CPython vs PyPy implementation), they could strip everything after the `-`, which will only be there for PyPy versions.

src/setup-python.ts
kurtmckee: I sorted it to help stabilize the cache key. If one developer writes their workflow YAML file as:

```yaml
- uses: actions/setup-python@v4
  with:
    python-version: |
      3.11
      3.10
      3.9
      3.8
      3.7
```

and the versions are later rearranged to:

```yaml
- uses: actions/setup-python@v4
  with:
    python-version: |
      3.7
      3.8
      3.9
      3.10
      3.11
```

sorting the versions will provide stability to the cache key.

kurtmckee:
@dmitry-shibanov I just saw the new support for "allow-prereleases", which is fantastic! I would love to get additional feedback about this PR, or to have it merged, so that I can use this new output as a cache key.

I don't know if I communicated the use case well, so to summarize:

Setting up a GitHub runner for each Python version is comparatively expensive for some of my projects -- it can take 30 seconds to spin up a runner, 5 seconds to set up the tox environment, but the test suite only takes 2 seconds to run for each Python version.

Therefore I install all Python versions on a single runner:

* 30 seconds to spin up the runner
* 5 seconds to restore the entire `.tox` directory -- for all Python versions -- from cache
* 2 seconds to run each Python version

The cache needs to be invalidated if _any_ Python version changes, and this PR makes that possible. Python 3.12 releases are coming out with frequency, and I would like to take advantage of the new "allow-prereleases" feature _and_ have this `python-versions` output available for cache busting.

Thanks for your time, and your work on setup-python!

kurtmckee:
@dmitry-shibanov, please let me know if you have additional feedback. I'm interested in having this PR merge for effective cache-busting when multiple Python versions are installed.

kurtmckee:
_Rebased on `main` and resolved CI failures._

kurtmckee:
I'm closing this PR because I've written a GitHub action to address my need for strong cache-busting based on installed Python versions.

https://github.com/kurtmckee/detect-pythons

PR #651, opened 2023-04-17
Use `@tsconfig/node16` instead of custom `tsconfig.json`

JamieMagee:
**Description:**
Use `@tsconfig/node16` instead of a custom `tsconfig.json`. Equivalent to:

```json
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "display": "Node 16",

  "compilerOptions": {
    "lib": ["es2021"],
    "module": "commonjs",
    "target": "es2021",

    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "moduleResolution": "node"
  }
}
```

Also required an updated version of TypeScript, to recognise `ES2021` as a valid target, and updated version of ncc to recognise the updated TypeScript.

**Related issue:**
N/A

**Check list:**

### DIFF
--- a/dist/cache-save/index.js
+++ b/dist/cache-save/index.js
@@ -59977,7 +59977,11 @@ module.exports = v4;
 
 var __createBinding = (this && this.__createBinding) || (Object.create ? (function(o, m, k, k2) {
     if (k2 === undefined) k2 = k;
-    Object.defineProperty(o, k2, { enumerable: true, get: function() { return m[k]; } });
+    var desc = Object.getOwnPropertyDescriptor(m, k);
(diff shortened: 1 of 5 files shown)


v-gowridurgad:
Hi @JamieMagee, thanks for the PR. At the time this was opened, this change made sense. Since then, the repository has moved to an ESM/NodeNext setup and the TypeScript/tooling configuration has changed quite a bit. Because of that, this PR no longer applies cleanly to the current project structure, and the related dependency and generated dist updates are now out of date as well. We’re going to close this out, but thank you again for the contribution.

Issue #627, opened 2023-03-15
Allow partial cache keys


toadjaune:
**Description:**

When restoring cache, the current cache management system only uses the primary cache key.

GitHub Actions cache allows specifying multiple prefixes, and restoring older caches, to speed up setup even when there has been a dependency change. Cf [corresponding documentation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key)

I'd recommend setting this as an opt-in behaviour, toggleable with a new input, because :
* That way, this would not be a breaking change
* Never rebuilding from scratch means stale data could stay forever in cache, which can be dangerous (e.g. a former dependency still present in a virtualenv)

Here's an example of what the configuration could look like (for poetry) : 

```yaml
    - uses: actions/setup-python@v4
      id: setup-python
      with:
        python-version: "3.10"
        cache: "poetry"
        allow-partial-restores: true
```

Here's a configuration roughly equivalent to what I'm proposing here (the cache key is slightly different, but the idea is the same)

```yaml
    - uses: actions/setup-python@v4
      id: setup-python
      with:
        python-version: "3.10"
    - uses: actions/cache@v3
      with:
        path: /home/runner/.cache/pypoetry/virtualenvs
        key: poetry-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('poetry.lock') }}
        restore-keys: |
          poetry-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('poetry.lock') }}
          poetry-${{ steps.setup-python.outputs.python-version }}-
```

We could also expose cache key management to the users through an input parameter, but that seems much more complex and brittle.

**Justification:**

This change can represent a significant performance improvement on lockfile change, easily several minutes, depending on the project.

**Are you willing to submit a PR?**
Not at the time, as I have a working workaround (presented above).

dmitry-shibanov:
Hello @toadjaune. Thank you for your report. We'll investigate the feature request.

dmitry-shibanov:
Hello @toadjaune. The purpose of caching is to cover the majority of use cases. If you need more customization we advise to use [actions/cache](https://github.com/actions/cache)

For now I'm going to close the issue.

tragiclifestories:
@dmitry-shibanov can this be reopened? Partial cache keys are not some obscure use case but an important optimisation widely used in the CI ecosystem. I would guess that 90+% of jobs that use setup-python would benefit from it.

[...]

PR #656, opened 2023-04-21
Allow partial cache keys for poetry environments

krishanbhasin-gc:
**Description:**
Allows the action to use partial-key matching as a fallback when the cache misses.

This will save much compute time for dependency upgrade scenarios.

**Related issue:**
https://github.com/actions/setup-python/issues/627

**Check list:**

### DIFF
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -300,7 +300,7 @@ virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/patrick/Library/Caches/py
             result = 'pipenv cache is not found';
             break;
           case 'poetry':
-            result = 'poetry cache is not found';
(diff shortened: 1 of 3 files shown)


dusan-trickovic:
Hello @krishanbhasin-gc ! I apologize for the late response. Unfortunately, our team has decided not to proceed with the original issue #627 and, therefore, this PR. The reason for this decision (as well as for closing the issue in the first place) mostly stems from the fact that this could cause the snowballing effect with dependencies - as the package manager can't handle the removal of stale ones. In other words, it would install new package versions, but would also keep the old ones.

With this in place, the cache size would continue to grow and grow until it, essentially, lost its original purpose and efficiency. So, at least right now, we don't think there's much of anything we can do regarding this.

In any case, however, we appreciate your input and thank you very much for your time and contribution!

Issue #693, opened 2023-06-27
Add GraalPy support


timfel:
**Description:**
I would like to add support for installing [GraalPy](https://github.com/oracle/graalpython) via setup-python. We are starting to ask people to include support for GraalPy in their packages, and they (reasonably) request to be able to test against GraalPy using their existing workflows using setup-python.

**Are you willing to submit a PR?**
I have prepared the changes for setup-python, mimicking the PyPy support code. The tests using local mock data pass.

timfel:
> Hello @timfel. Thank you for your feature request. We'll investigate it.

Thanks. I opened a PR to implement this at https://github.com/actions/setup-python/pull/694

dmitry-shibanov:
Hello @timfel. We've merged the pull request. Could you please try to use `actions/setup-python@main` and confirm that it works as expected for you. For now I'm going to reopen the issue until the new version of action is released.

kurtmckee:
I tested this using the following configuration:

```yaml
- uses: "actions/setup-python@v5"
  with:
    python-version: |
      graalpy-23.1
      pypy-3.10
      3.11
```

[Link to run, which failed on Windows](https://github.com/kurtmckee/detect-pythons/actions/runs/7169923447/job/19521432415?pr=12#step:2:14)

For Linux and macOS, `graalpy-23.1` resolved to version [23.1.1](https://github.com/oracle/graalpython/releases/tag/graal-23.1.1).

However, on Windows, [`setup-python` reported that no "23.1" version existed](https://github.com/kurtmckee/detect-pythons/actions/runs/7169923447/job/19521432415?pr=12#step:2:14):

```
Run actions/setup-python@v5
  with:
    python-version: graalpy-23.1
  pypy-3.10
  3.11
  
    check-latest: false
    token: ***
    update-environment: true
    allow-prereleases: false
Installed versions
  GraalPy version 23.1 was not found in the local cache
  Error: GraalPy version 23.1 with arch x64 not found
```

This is the complete list of files for the GraalPy 23.1.1 release at the time of writing:

```
graalpy-23.1.1-linux-aarch64.tar.gz
graalpy-23.1.1-linux-aarch64.tar.gz.sha256
graalpy-23.1.1-linux-amd64.tar.gz
graalpy-23.1.1-linux-amd64.tar.gz.sha256
graalpy-23.1.1-macos-aarch64.tar.gz
graalpy-23.1.1-macos-aarch64.tar.gz.sha256
[...]

PR #694, opened 2023-06-28
Add GraalPy support

timfel:
**Description:**
This adds support for graalpy as a Python to use when running setup-python

**Related issue:**
https://github.com/actions/setup-python/issues/693

**Check list:**


timfel:
Locally `npm test` is passing and running the `test-graalpy` workflow I added with `act` also passes.

timfel:
Any thoughts?

dmitry-shibanov:
Hello @timfel. Sorry for the late response. Could you please attach licenses information regarding to GraalPy ?

timfel:
> Hello @timfel. Sorry for the late response. Could you please attach licenses information regarding to GraalPy ?

Hi. My changes pull GraalPy from the open source Github repo, which is under UPL: https://github.com/oracle/graalpython/blob/master/LICENSE.txt#L5

timfel:
Anything else I can help with?

dmitry-shibanov:
Hello @timfel. Sorry for the late response. For now we're investigating it. We'll ping you If we need any help. Thank you.

src/install-graalpy.ts
```diff
@@ -0,0 +1,256 @@
...
+      }
+      if (err.stack !== undefined) {
+        core.debug(err.stack);
+      }
+    }
+    throw err;
+  }
+}
+
+export async function getAvailableGraalPyVersions() {
```
dmitry-shibanov, line 103: I think we need to add extra logic because through [GitHub API releases](https://docs.github.com/en/free-pro-team@latest/rest/releases/releases?apiVersion=2022-11-28#list-releases) we can get only 100 releases per page (by 30 per page). You need to specify releases per page and go through all page

src/install-graalpy.ts
```diff
@@ -0,0 +1,256 @@
...
+  const url = 'https://api.github.com/repos/oracle/graalpython/releases';
+  const http: httpm.HttpClient = new httpm.HttpClient('tool-cache');
```
dmitry-shibanov, line 105: The [token](https://github.com/actions/setup-python/blob/main/action.yml#L18-L20) needs to be added for authorization to avoid rate limit.

src/install-graalpy.ts
```diff
@@ -0,0 +1,256 @@
...
+  platform: string
+) {
+  const graalpyArch =
+    architecture === 'x64'
+      ? 'amd64'
+      : architecture === 'arm64'
+      ? 'aarch64'
+      : architecture;
+  const graalpyPlatform =
+    platform === 'win32'
```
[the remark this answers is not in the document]
msimacek: `platform` here is a parameter to the function. I think it would be weird if it overrode the passed parameter with the current platform.

msimacek:
Hi @dmitry-shibanov, I believe I addressed all your comments (except one where I commented). Could you please check again?

.github/workflows/test-graalpy.yml
```diff
@@ -0,0 +1,116 @@
...
+      - '**.md'
+
+jobs:
+  setup-graalpy:
+    name: Setup GraalPy ${{ matrix.graalpy }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, ubuntu-20.04, ubuntu-latest]
```
kurtmckee, line 20: GraalPy is available for Windows, too. Would it be an option to test against Windows?

__tests__/find-graalpy.test.ts
```diff
@@ -0,0 +1,378 @@
...
+import {IGraalPyManifestRelease, IS_WINDOWS} from '../src/utils';
+
+import manifestData from './data/graalpy.json';
+
+const architecture = 'x64';
+
+const toolDir = path.join(__dirname, 'runner', 'tools');
+const tempDir = path.join(__dirname, 'runner', 'temp');
+
+/* GraalPy doesn't have a windows release yet */
```
kurtmckee, line 22: GraalPy is downloadable for Windows; is this comment still correct?

.github/workflows/test-graalpy.yml
timfel: For 23.0 there were no Windows binary releases. The upcoming 23.1 will have Windows releases, but they are very limited, things like native extensions do not work, yet. Tests for Windows will likely just not pass.

__tests__/find-graalpy.test.ts
timfel: It is correct, the last release for GraalPy is 23.0.1 and doesn't have Windows binaries. The upcoming 23.1 will have binaries, but they are very limited.

.github/workflows/test-graalpy.yml
timfel: Our plan is to have more complete Windows support for our 2024 release, so we'll have to do a follow up PR next year.

timfel:
Anything else to do here? We would really like this to go in so that Python projects that already use setup-python can easily start testing against GraalPy :)

timfel:
Friendly ping :)

dmitry-shibanov (approved):

kurtmckee:
When this merges, I'll work to update #611 so that GraalPy is supported for robust cache-busting as well.

kurtmckee:
@timfel It looks like the 24.x release series is out. Does this include Windows support that could integrate with setup-python? :crossed_fingers:

timfel:
> @timfel It looks like the 24.x release series is out. Does this include Windows support that could integrate with setup-python? 🤞

It does indeed. We also finally managed to get nightly builds out with some json metadata (https://github.com/graalvm/graal-languages-ea-builds/blob/main/graalpy/versions/latest-ea.json), so we'll integrate that, too, so people can test on upcoming versions as well.

Issue #683, opened 2023-06-14
Unable to use built-in GitHub Token for avoiding rate limit issues on GHES


mzwiesler:
**Description:**
For our enterprise we are facing "API rate limit exceeded ..." errors. I read the documentation how to [avoid rate limit issues](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#avoiding-rate-limit-issues) but for our enterprise using PAT is not feasible. I would like to use the `GITHUB-TOKEN` for [authentication in my Github Workflow](https://docs.github.com/en/rest/overview/authenticating-to-the-rest-api?apiVersion=2022-11-28#authenticating-in-a-github-actions-workflow) since this is the recommended way from GitHub. However, this is not possible at the moment resulting in a `Bad Credentials` error.

It seems that the [AUTH variable](https://github.com/actions/setup-python/blob/bd6b4b6205c4dbad673328db7b31b7fab9e241c0/src/install-python.ts#L9) variable is not usable in this case since only personal tokens are authenticated with `Authorization: token <TOKEN>` but for `GITHUB-TOKEN` one needs to use `Authorization: <GITHUB_TOKEN>` instead as can be seen in the workflow below in step `read limit`.

**Action version:**
4.6.1

**Platform:**
- [x] Ubuntu

**Runner type:**
- [x] Self-hosted

**Tools version:**

All

**Repro steps:**  

On GHES, following [this example](https://github.com/actions/setup-python/pull/443#issuecomment-1206776401) I used the following workflow:

```
name: CI With GitHub Token

on: [workflow_dispatch]

jobs:
  sample_build:
    runs-on: [azure-runners]
    steps:
      - name: "rate limit"
        run: |
          curl \
            -H "Accept: application/vnd.github.v3+json" \
            -H "Authorization: $TOKEN" \
            https://api.github.com/rate_limit
        env:
          TOKEN: ${{ secrets.GITHUB_TOKEN }}

      - name: Clear the tool cache
        run: rm -rf /Users/runner/hostedtoolcache/Python/

      - name: Set up Python
        uses: actions/setup-python@v4.6.1
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          python-version: 3.9.12

      - name: "rate limit"
        run: |
          curl \
            -H "Accept: application/vnd.github.v3+json" \
            -H "Authorization: $TOKEN" \
            https://api.github.com/rate_limit
        env:
[...]

Issue #666, opened 2023-05-11
Getting Bad Credentials error when passing token input in GHE


alvaro-ajv:
**Description:**
When trying to pass the token input to reduce rate limit error, we are getting "Bad Credentials" error; using previous versions of the action it works, but using the last release we get that error. 

Seems that the bug was introduced on these PR: [443](https://github.com/actions/setup-python/pull/443)
Also the same change from the mentioned PR was tested and reverted in these PRs respectively: [427](https://github.com/actions/setup-python/pull/427) and [437](https://github.com/actions/setup-python/pull/437)

Action version 4.2.0 that is before of the change works fine.

We are trying to use the latest version of the action due to deprecation of "set-output" command as described in the [docs](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), and as is mentioned in PR [578](https://github.com/actions/setup-python/issues/578), latest version should remove the warning, but we cannot use later versions than 4.2.0 due to above bug.

Bad credentials error:

![image](https://github.com/actions/setup-python/assets/40005341/47bbc20d-230b-4a2f-bc21-46a885ac769d)

"set-ouput" warning:

![image](https://github.com/actions/setup-python/assets/40005341/c767649b-84f7-4ebf-b4a4-dce8f6d606ff)

 

**Action version:**
Latest

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [ ] Hosted
- [x] Self-hosted

IvanZosimov:
Hi, @alvaro-ajv 👋 Thanks for the issue, we will take a look!

KFo3:
Similar to Disney's error might be the same bug.  Please correct me if I am wrong.
[ZD Ticket](https://github.zendesk.com/agent/tickets/2169251) 

 Disney just started experiencing this problem after updating GHES to 3.8.3.  They are also utilizing PAT over tokens.  Self hosted runners using  `ubuntu:jammy` 22.04.2

Current runner version: '2.304.0'

```
Run actions/setup-python@v4
Version 3.10 was not found in the local cache
Error: Bad credentials
```

priya-kinthali:
Hello @alvaro-ajv 👋,
Could you kindly verify if the issue related to the rate limit has been resolved and confirm that no other issues have emerged while using the [setup-actions-demo/setup-python@main branch](https://github.com/setup-actions-demo/setup-python).

priya-kinthali:
[...]

Issue #638, opened 2023-03-28
Rate limit token does not work on GHE


mvs5465:
**Description:**
Using the `token` parameter for the Github Action, the token is not used. 

**Action version:**
v4

**Platform:**
- [x] Ubuntu

**Runner type:**
- [x] Self-hosted

**Tools version:**
3.x

**Repro steps:**  
Add a Github Action workflow like this to your Github Enterprise repository:

```
      - name: Install Python
        uses: actions/setup-python@v4
        with:
          token: ${{ secrets.DOTCOM_TOKEN }}
          python-version: '3.x'
```

Then also add https://github.com/actions/setup-python/pull/510#discussion_r982394709 using the same token variable.

We can tell by the rate-limit response that the token has not been used. 

**Expected behavior:**

Expect action to use token.

**Actual behavior:**

`API rate limit exceeded for <ip>`

We confirmed using the rate limit query that the rate limit quotas have not changed, and thus the token has not been used by the action.

e-korolevskii:
Hello @mvs5465 
Thank you for your report. We'll investigate the issue and get back to you.

jay0129:
I got the same problem.
I downgraded action version to `actions/setup-python@v3.1.2` and it successfully uses `token`.

chebong-aus:
Hi @e-korolevskii, are there any updates on this issue? I am facing the exact same issue in my GitHub Enterprise repository.

dmitry-shibanov:
Hello everyone. Could you please take a look at this [issue](https://github.com/actions/setup-python/issues/683) because I think they're related.

chebong-aus:
Hi @dmitry-shibanov, looking at the link you provided, not sure if they are related because the error I get is `Error: API rate limit exceeded for x. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)`. This is while passing in a GitHub.com Token.

priya-kinthali:
Hello @mvs5465 👋,
Could you kindly verify if the issue related to the rate limit has been resolved and confirm that no other issues have emerged while using the [setup-actions-demo/setup-python@main branch](https://github.com/setup-actions-demo/setup-python).

priya-kinthali:
Hello @mvs5465👋,
Just a gentle reminder, Could you kindly verify if the issue related to the rate limit has been resolved and confirm that no other issues have emerged while using the [setup-actions-demo/setup-python@main branch](https://github.com/setup-actions-demo/setup-python).

[...]

PR #443, opened 2022-06-27
Only use github.token on github.com

ChristopherHX:
**Description:**
This expression `${{ github.server_url == 'https://github.com' && github.token || '' }}` evaluates to `''` if called from GHES hosted elsewhere
You can still provide your token on both github.com and GHES
See docu https://docs.github.com/en/actions/learn-github-actions/contexts.

**Related issue:**
Needs PullRequest: #427
#229, without the need for a new input or breaking using the token input on GHES.

**Check list:**
- [ ] Mark if documentation changes are required.
- [ ] Mark if tests were added or updated to cover the changes.

PR #720, opened 2023-08-23
Add support for locally hosted GHES instances to reduce rate limiting

C1ARKGABLE:
**Description:**
Rate limiting, while necessary, can be frustrating to deal with. The [work around](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#avoiding-rate-limit-issues) is not always a viable solution. No company wants to rely solely on a single employee's GitHub.com PAT for significant internal projects.
This attempts to resolve rate limiting errors by allowing users to internally host a copy of [`actions/python-versions`](https://github.com/actions/python-versions) and specify their GHES api endpoint instead of the GitHub.com endpoint being hardcoded. 

`##[error]API rate limit exceeded for YOUR_IP. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)`

**Related issue:**
https://github.com/actions/setup-python/issues/683
https://github.com/actions/setup-python/issues/638
https://github.com/actions/setup-python/issues/666
https://github.com/actions/setup-python/pull/316

https://github.com/actions/setup-python/pull/443#issuecomment-1206776401

https://docs.github.com/en/enterprise-server/admin/github-actions/managing-access-to-actions-from-githubcom/setting-up-the-tool-cache-on-self-hosted-runners-without-internet-access

**Check list:**

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -10957,10 +10957,10 @@ function findAllVersions(toolName, arch) {
     return versions;
 }
 exports.findAllVersions = findAllVersions;
-function getManifestFromRepo(owner, repo, auth, branch = 'master') {
+function getManifestFromRepo(owner, repo, auth, branch = 'master', serverUrl = "https://api.github.com") {
     return __awaiter(this, void 0, void 0, function* () {
         let releases = [];
-        const treeUrl = `https://api.github.com/repos/${owner}/${repo}/git/trees/${branch}`;
+        const treeUrl = `${serverUrl}/repos/${owner}/${repo}/git/trees/${branch}`;
         const http = new httpm.HttpClient('tool-cache');
         const headers = {};
         if (auth) {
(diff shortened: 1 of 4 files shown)


C1ARKGABLE:
Guidance + suggestions are welcome!

bhundven:
@e-korolevskii @dmitry-shibanov This also affects a GHE instance I use.

dist/setup/index.js
```diff
@@ -10957,10 +10957,10 @@ function findAllVersions(toolName, arch) {
     return versions;
 }
 exports.findAllVersions = findAllVersions;
-function getManifestFromRepo(owner, repo, auth, branch = 'master') {
+function getManifestFromRepo(owner, repo, auth, branch = 'master', serverUrl = 'https://api.github.com') {
```
jhaxon, line 10960: `getManifestFromRepo()` is a library function [imported at the top of `src/install-python.ts`](https://github.com/actions/setup-python/blob/23e715c78e7c4870b8437b7105eeb5f54ad640f0/src/install-python.ts#L3). The code here is generated from that library at compile time, so your changes will be overwritten.

In order for this PR to work you need to limit your changes to the `src/*.ts` files.

src/install-python.ts
```diff
@@ -10,7 +10,11 @@ const AUTH = !TOKEN ? undefined : `token ${TOKEN}`;
 const MANIFEST_REPO_OWNER = 'actions';
 const MANIFEST_REPO_NAME = 'python-versions';
 const MANIFEST_REPO_BRANCH = 'main';
-export const MANIFEST_URL = `https://raw.githubusercontent.com/${MANIFEST_REPO_OWNER}/${MANIFEST_REPO_NAME}/${MANIFEST_REPO_BRANCH}/versions-manifest.json`;
+const API_URL = core.getInput('github_api_url');
+const GITHUB_API_URL = API_URL ? 'https://api.github.com' : API_URL;
```
jhaxon, line 14: truthy and falsey expressions are switched here and line 16

src/install-python.ts
```diff
@@ -10,7 +10,11 @@ const AUTH = !TOKEN ? undefined : `token ${TOKEN}`;
...
+const RAW_URL = core.getInput('github_raw_url');
+const GITHUB_RAW_URL = RAW_URL ? 'https://raw.githubusercontent.com' : RAW_URL;
```
jhaxon, line 16: truthy and falsey expressions are switched

src/install-python.ts
C1ARKGABLE: Good catch

jhaxon: The unit tests caught this! ;)

jhaxon:
Without any changes to `installCpythonFromRelease` the action will download the release from `github.com`. Further work is needed here.

C1ARKGABLE:
Converted to a draft until I can work on this further

v-priyagupta108:
@C1ARKGABLE, 
Thanks for the PR! We’ve since merged and released #766, which addressed the GHES rate-limiting problem via the raw-URL fallback.
Could you clarify whether there’s still a gap or issue that this PR would address?

v-priya-kinthali:
Hello @C1ARKGABLE 👋, As we didn't hear back on whether there's a remaining gap this PR would still cover, and given that the underlying problem is now resolved on main, we're going to close this PR. Thanks again for your contribution!

PR #728, opened 2023-09-08
perf: add dynamically loaded cache-dependencies module

aminya:
**Description:**
This reduces the main bundle size by ⚠️  1.1 MB minified! This bundle is loaded only when caching is enabled.

The reason for this huge reduction is that caching dependencies uses the `@actions/cache` package, which is an extremely large package with big dependencies.

`setup-python` is used in `setup-cpp` as a library. This optimization reduces the bundle size for that package as well.

**Related issue:**

**Check list:**


aminya:
@konradpabjan could you review this?

PR #766, opened 2023-11-15
feat: fallback to "raw" endpoint for manifest when rate limit is reached

### LINKED ISSUE #756: Enhance setup-python Action in GitHub Enterprise Server with Raw API Fallback Mechanism
**Description:**

In the context of GitHub Enterprise Server (GHES), it is typical for runners to share the same IP address due to Network Address Translation (NAT). This configuration results in the quick exhaustion of the unauthenticated rate limit (60 requests per hour per IP address) when accessing the `versions-manifest.json` file, leading to failures in the `setup-python` action. The current workaround is to incorporate [a "github.com" token into the `setup-python` action](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#avoiding-rate-limit-issues). Although this solution is functional, it necessitates the creation of an additional github.com token (and technical user) for each repository/team, which is not optimal. Moreover, Pull Requests from forks are not supported due to their lack of access to the secret. 

A potential improvement could be leveraging the `raw` API to retrieve the version-manifest, as it does not impose a rate limit and hence facilitates unrestricted consumption without the need for a token.

**Justification:**
Our GitHub Enterprise Server has frequently encountered rate limit issues, and the manual token addition method does not scale well. Utilizing the raw API as a fallback could provide an automatic workaround for this issue. If this approach fails, users can still resort to adding a token.

I have created a fork to verify that the raw API isn't affected by the rate limit and can be used as a fallback solution. (https://github.com/actions/setup-python/compare/main...Shegox:setup-python:raw-access?expand=1)

The log extract from the workaround is as follows:

```
  Version 3.10 was not found in the local cache
  ##[debug]Getting manifest from actions/python-versions@main
  ##[debug]{"name":"HttpClientError","statusCode":403,"result":{"message":"API rate limit exceeded for xx.xx.xx.xx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}}
  ##[debug]Fetching via the API failed. Fetching using raw URL.
  ##[debug]check 3.13.0-alpha.1 satisfies 3.10
  ##[debug]check 3.12.0 satisfies 3.10
  ##[debug]check 3.12.0-rc.3 satisfies 3.10
  ##[debug]check 3.12.0-rc.2 satisfies 3.10
  ##[debug]check 3.12.0-rc.1 satisfies 3.10
```

I am prepared to submit a PR incorporating the proposed changes. 
Before proposing this PR I wanted to check if that would be accepted or if there are other ideas to overcome this?

**Related Issues**

- https://github.com/actions/setup-python/issues/683
- https://github.com/actions/setup-python/issues/638
- https://github.com/actions/setup-python/issues/666
- https://github.com/actions/setup-python/pull/316
- #720

Shegox:
**Description:**
This PR leverages the `raw` endpoint (`https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json`) to retrieve the python manifest as a fallback mechanism. This allows GitHub Actions Runners, which have no API token and exceed the anonymous rate limit of 60 requests/hour/IP to still retrieve the manifest list without needing credentials. 
This is especially helpful for GitHub Enterprise Server and self-hosted runners, as there multiple runners might share the same IP-address and quickly exceed the available rate limit. 
The implementation as a fallback mechanism as well minimizes the risk of breaking current workflows, as they will still use the normal API to retrieve the manifest.  

**Related issue:**
Fixes #756 

**Check list:**


winlover32 (approved):
chek

aparnajyothi-y (approved):

priya-kinthali (approved):

HarithaVattikuti (approved):

Issue #774, opened 2023-12-06
`python-version-file` chooses most recent supported Python version, not the oldest supported version


ErikBjare:
**Description:**

I'm using `python-version-file: pyproject.toml` to specify which Python version to use. My `pyproject.toml` is pretty standard poetry config, with a `python = "^3.8"` marker.

However, when using the action with such a config, it uses the *most recent* supported Python version. This doesn't make much sense to me, as Python code is generally forwards-compatible (except numpy/scipy/etc), but it is not backwards-compatible, as there are features in newer versions that aren't supported in previous versions (which is why I always run my CI jobs with the lowest supported Python minor version).

I expected it to pick the lowest supported minor version with the highest patch. Such that `^3.8` picks `3.8.15`, not `3.12.0`. 

**Justification:**

With the current behavior, using `python-version-file` with a `pyproject.toml` that has a Python version marker specified with `^` (common) would basically be a no-op, as it will pick the same version as if `python-version-file` isn't specified at all.

Thus, I don't see much use for the current behavior, and would like to see it changed.

One option is to add a config to choose the behavior, which would prevent breakage in people's existing workflows. (as I did in https://github.com/actions/setup-python/pull/775)

**Are you willing to submit a PR?**

Yes! Done in https://github.com/actions/setup-python/pull/775

ErikBjare:
@dmitry-shibanov Did you see my PR?

The current behavior really makes the `python-version-file` kinda worthless for use with Poetry `pyproject.toml`, for the reasons previously stated.

I'm aware I can just set the `python-version` like that, as that's what I've been doing, but I'd like to keep the Python version specified in a single place (as `python-version-file` enables).

I'd appreciate if you could give this a second consideration and reopen the issue, as this really surprised me and the current behavior is (imo) not usable.

It's a small and non-breaking change to adopt the additional configuration added in the PR.

dmitry-shibanov:
It is an expected behaviour from the poetry side that they use caret notation. In this case [poetry follows SemVer notation](https://github.com/python-poetry/poetry/issues/3427#issuecomment-1001973463) format that is why it is an expected behaviour to take the latest minor version. Caret and tilde are both valid variants for setup-python and poetry but they take different ranges. The action should not change the actual supported format from poetry through another input. You can try to use `poetry env use <version>` command, python-version input or change caret to tilde directly in the toml file.

Andrej730:
Also got the same idea.

[...]

PR #775, opened 2023-12-06
feat: added python-version-strategy config option, to enable enforcing use of the lowest supported (approximate) version

ErikBjare:
**Description:**

Adds a `python-version-strategy` config option that enables overriding the version-picking strategy provided by the Python version specifier from `>=` and `^` to `~`.

This is very useful when the version specified in `python-version-file` is a version constraint such as `^3.8` but where you usually want to test against the lowest supported version (with the highest patch version).

**Related issue:**

Filed a feature request for this here: #774 

**Check list:**

   - Not done, yet (will do if PR is likely to be accepted)
   - Not done, yet (will do if PR is likely to be accepted)

**PRs I will merge if this change is accepted:**

 - https://github.com/ActivityWatch/aw-core/pull/124
 - https://github.com/ActivityWatch/aw-client/pull/80
 - https://github.com/ActivityWatch/aw-qt/pull/101
 - https://github.com/ActivityWatch/aw-server/pull/148
 - https://github.com/ActivityWatch/aw-watcher-afk/pull/65
 - https://github.com/ActivityWatch/aw-watcher-window/pull/93
 - https://github.com/ActivityWatch/aw-watcher-input/pull/29

### DIFF
--- a/src/setup-python.ts
+++ b/src/setup-python.ts
@@ -53,6 +53,7 @@ function resolveVersionInputFromDefaultFile(): string[] {
 function resolveVersionInput() {
   let versions = core.getMultilineInput('python-version');
   const versionFile = core.getInput('python-version-file');
+  const versionStrategy = core.getInput('python-version-strategy');
 
   if (versions.length) {
     if (versionFile) {
@@ -72,6 +73,9 @@ function resolveVersionInput() {
       versions = resolveVersionInputFromDefaultFile();
     }
   }
+
+  if(versionStrategy == "approximate")
+    versions.map(v => v.replace(">=", "~").replace("^", "~"));
 
   return versions;
 }


winlover32 (approved):

src/setup-python.ts
```diff
@@ -73,6 +74,9 @@ function resolveVersionInput() {
     }
   }
 
+  if(versionStrategy == "approximate")
```
ErikBjare, line 77: Might be better to make the `versionStrategy` directly take `~` as an argument, so that it can be used for all kinds of version overrides.

Not sure if there are uses for overriding the strategy to `^` however.

winlover32 (approved):

v-gowridurgad:
Hi @ErikBjare, Thanks for the contribution.

The behavior this PR aims to override follows the action’s existing SemVer-based version resolution: a caret range like `^3.8` is treated as a compatible range and can resolve to the newest satisfying version within it, whether it comes from `python-version` or `python-version-file`. Reinterpreting that inside the action would change the semantics of the version declared in the source file rather than simply reading it as written.

Users who want minimum-compatible behavior can express that directly with a narrower specifier such as `~3.8` or `3.8`.

Since the underlying feature request in #774 was closed as not planned on that basis, this PR does not have a path to merge. Closing this out accordingly.

Issue #819, opened 2024-02-19
Failed to unzip the Python package  on Windows-2022 Core server 


imarinov3:
**Description:**

**Action version:**
2.313.0

**Platform:**
- [ ] Ubuntu
- [ ] macOS
- [x] Windows

**Runner type:**
- [ ] Hosted
- [x] Self-hosted

**Tools version:**

**Repro steps:**  
``` yml
 - name: Setup python
        uses: actions/setup-python@v5
        with:
          python-version: '3.6.x - 3.11.x'
```

**Expected behavior:**
To install the required Python version

**Actual behavior:**
Failed to unzip Python package 

Run actions/setup-python@v5
  with:
    python-version: 3.6.x - 3.11.x
    check-latest: false
    token: ***
    update-environment: true
    allow-prereleases: false
  env:
    DEBUG: 1
  
Installed versions
  Version 3.6.x - 3.11.x was not found in the local cache
  Version 3.6.x - 3.11.x is available for downloading
  Download from "https://github.com/actions/python-versions/releases/download/3.11.8-7809691605/python-3.11.8-win[32](https://github.com/abaltatech/weblink/actions/runs/7950776332/job/21720913455#step:3:34)-x64.zip"
  Extract downloaded archive
[...]

PR #827, opened 2024-02-29
FIX: Extract zip win

SMoraisAnsys:
**Description:**

Changes consist in reordering extract utils to first attemps ExtractToDirectory and then Expand-Archive. For some reason the download CPython release was not correctly handled by Expand-Archive. This could be happening if the zip file was compressed using an unsupported compression method (e.g. through `zip` with Unix/Linux).

**Related issue:** https://github.com/actions/setup-python/issues/819

**Check list:**

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -9877,11 +9877,12 @@ function extractZipWin(file, dest) {
             yield exec_1.exec(`"${pwshPath}"`, args);
         }
         else {
+            //attempt to use pwsh with ExtractToDirectory, if this fails attempt Expand-Archive
             const powershellCommand = [
                 `$ErrorActionPreference = 'Stop' ;`,
                 `try { Add-Type -AssemblyName System.IO.Compression.FileSystem } catch { } ;`,
-                `if ((Get-Command -Name Expand-Archive -Module Microsoft.PowerShell.Archive -ErrorAction Ignore)) { Expand-Archive -LiteralPath '${escapedFile}' -DestinationPath '${escapedDest}' -Force }`,
-                `else {[System.IO.Compression.ZipFile]::ExtractToDirectory('${escapedFile}', '${escapedDest}', $true) }`
+                `try { [System.IO.Compression.ZipFile]::ExtractToDirectory('${escapedFile}', '${escapedDest}', $true) }`,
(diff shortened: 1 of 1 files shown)


SMoraisAnsys:
@priyagupta108 I took the liberty to take a look at #819 and I think this changes should correct the issue. I tested it on my repo and it works fine (didn't do intensive testing though).

RobPasMue (approved):
This will solve so many issues! Windows self-hosted runners could no longer use "v5" if the Python version requested was not available in the cache or in the system already. We can finally use the action as is without having to stick to "v4" or implement workarounds!

SMoraisAnsys:
I took the liberty to reproduce the error (fixed by this PR) on a github runner. The associated branch (https://github.com/SMoraisAnsys/setup-python/tree/test/no-pwshpath) consists in removing a part of the code of  `function extractZipWin(file, dest)` in order to ensure that the runner uses the failing part of the code.

PR that can be used to oversee the changes : https://github.com/SMoraisAnsys/setup-python/pull/1
Failing CI log : https://github.com/SMoraisAnsys/setup-python/actions/runs/8491166206/job/23262896827?pr=1

RobPasMue:
@actions/setup-actions-team could we get this reviewed and merged soon?

jeremyd2019:
@aparnajyothi-y pointed out in https://github.com/actions/setup-python/issues/819#issuecomment-2100107374 that this code comes from actions/toolkit.  Perhaps a PR against https://github.com/actions/toolkit/blob/main/packages/tool-cache/src/tool-cache.ts would be better reviewed?

SMoraisAnsys:
> @aparnajyothi-y pointed out in [#819 (comment)](https://github.com/actions/setup-python/issues/819#issuecomment-2100107374) that this code comes from actions/toolkit. Perhaps a PR against https://github.com/actions/toolkit/blob/main/packages/tool-cache/src/tool-cache.ts would be better reviewed?

As you mentioned, the problem seems to be related with `actions/toolkit`. However, the mentioned [PR](https://github.com/actions/toolkit/pull/1552) was created in October 2023. I don't think changes would be better reviewed there.

jeremyd2019:
I hear you, but I think what I meant is that the changes need to be applied there rather than here.  Also, I don't think that linked PR would actually help this case.

SMoraisAnsys:
Closing this since #916 fixed #819 in our CICD.

Issue #585, opened 2023-01-11
Unable to Set-up Python


Pavansai07784:
**Description:**
Version 3.10 was not found in the local cache
Error: Version 3.10 with arch x[6]

**Action version:**
Latest

**Platform:**
- Ubuntu

**Runner type:**
-  Self-hosted

**Tools version:**
3.10, I have also tried 3.10.4 I get the same error

**Repro steps:**  
     - name: Set up Python 3.10
        uses: actions/setup-python@v3
        with:
          python-version: "3.10"

**Expected behavior:**
Version should be fetched correctly.

**Actual behavior:**
The 3.10 version for x64 exists but the error says otherwise.

dmitry-shibanov:
Hello @Pavansai07784. Thank you for your report. Which version of ubuntu do you use ? Where are your binaries located ? Did you define `RUNNER_TOOL_CACHE` or `AGENT_TOOLSDIRECTORY` env variables? Which values do they have ?

Pavansai07784:
Hi, @dmitry-shibanov Thanks for responding quickly, It is a 64-bit Linux VM, and what do you mean by Binaries here? I haven't specified or RUNNER_TOOL_CACHE or AGENT_TOOLSDIRECTORY

smithsp:
I am seeing a similar error.  
yaml at https://github.com/gafusion/omas/blob/ec_cocos/.github/workflows/regression.yml 
Error at https://github.com/gafusion/omas/actions/runs/3896178260  

Mention @orso82 and @jmcclena

dmitry-shibanov:
> Hi, @dmitry-shibanov Thanks for responding quickly, It is a 64-bit Linux VM, and what do you mean by Binaries here? I haven't specified RUNNER_TOOL_CACHE or AGENT_TOOLSDIRECTORY

I mean where do you store python versions ? Could you please try to log values for RUNNER_TOOL_CACHE and AGENT_TOOLSDIRECTORY ?

Pavansai07784:
This is what i am using    
  - name: Set up Python 3.8
        uses: actions/setup-python@v4
        with:
          python-version: "3.8.16"
          architecture: x64
        env:
          AGENT_TOOLSDIRECTORY: /opt/github/actions-runner/_work/cache

dmitry-shibanov:
[...]

PR #933, opened 2024-08-29
feat: add ignore-platform-version option to allow install on debian

loopingz:
**Description:**
When trying to use `setup-python` on a Debian-based runner, an error is thrown that it cannot find the python SDK.

```
Error: The version '3.12.3' with architecture 'x64' was not found for Debian 11.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```
It comes down to the fast that the version manifest contains Ubuntu versions number that are irrelevant to Debian, I'm unsure how relevant the platform_version is compare to architecture and platform, because the downloaded file are linux binaries not specific to a distribution I think.

Something must have changed because this is a new behavior.

**Related issue:**
[Add link to the related issue.](https://github.com/actions/setup-python/issues/585)

**Check list:**


aparnajyothi-y:
Hello @loopingz , Thank you for the pull request. While adding an ignore-platform-version option would enhance compatibility for Debian-based runners, it introduces additional complexity and could lead to potential compatibility issues, especially if users install versions that are not fully supported by their OS. As per our [support policy](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-setup-python-with-a-self-hosted-runner), we only support versions available on our hosted runners. This ensures a more consistent and reliable experience across all environments.

We appreciate your understanding and encourage you to continue sharing your feedback with us.

aparnajyothi-y:
Hello @loopingz, Please let us know if you have any concerns or need further clarifications regarding the above comment.

aparnajyothi-y:
Hello @loopingz, Please let us know if you have any concerns or need further clarifications regarding the above comment.

Issue #436, opened 2022-06-14
Post setup Python step fails (pipenv cache)


mrlubos:
**Description:**

Hey! So this used to work just a few days ago. We’ve been having some issues with cache in the last few days (see [here for details](https://stackoverflow.com/questions/72598852/getcacheentry-failed-cache-service-responded-with-503)), this is a brand new issue which I think might or might not be related to the previous problem. Simply put, the action fails in the post step with the following message.

```sh
Post job cleanup.
[2](https://github.com/<org>/<repo>/runs/6889538635?check_suite_focus=true#step:20:2)
Error: Cache folder path is retrieved for pipenv but doesn't exist on disk: /home/runner/.local/share/virtualenvs
```

Here’s the actual step:

```sh
Run actions/setup-python@v3.1.2
[2](https://github.com/<org>/<repo>/runs/6889538635?check_suite_focus=true#step:5:2)
  with:
[3](https://github.com/<org>/<repo>/runs/6889538635?check_suite_focus=true#step:5:3)
    cache: pipenv
[4](https://github.com/<org>/<repo>/runs/6889538635?check_suite_focus=true#step:5:4)
    python-version: 3.9.13
[5](https://github.com/<org>/<repo>/runs/6889538635?check_suite_focus=true#step:5:5)
    token: ***
[6](https://github.com/<org>/<repo>/runs/6889538635?check_suite_focus=true#step:5:7)
Successfully setup CPython (3.9.13)
[7](https://github.com/<org>/<repo>/runs/6889538635?check_suite_focus=true#step:5:8)
pipenv cache is not found
```

**Action version:**
Specify the action version

3.1.2

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**

3.9.13

**Repro steps:**  
A description with steps to reproduce the issue. If your have a public example or repo to share, please provide the link.

**Expected behavior:**
A description of what you expected to happen.

No issue occurs, the same way it used to work a week ago.

**Actual behavior:**
A description of what is actually happening.

This error message appears on every run.

dmitry-shibanov:
Hello @mrlubos. Thank you for your report. I've tried to reproduce the issue but everything works as expected. Could you please provide public repository to reproduce the issue ?

mrlubos:
[...]

PR #968, opened 2024-10-23
Update cache error message when no dependencies are found to cache for improved clarity

aparnajyothi-y:
Description:
This PR updates the error messag for pip and pipenv to provide clearer information when the cache folder path is retrieved but does not exist on disk. This enhancement aims to improve troubleshooting by specifying the package manager and the exact paths being checked.

Related issue:
https://github.com/actions/setup-python/issues/436

### DIFF
--- a/src/cache-save.ts
+++ b/src/cache-save.ts
@@ -41,7 +41,7 @@ async function saveCache(packageManager: string) {
     throw new Error(
       `Cache folder path is retrieved for ${packageManager} but doesn't exist on disk: ${cachePaths.join(
         ', '
-      )}`
+      )}. This likely indicates that there are no dependencies to cache. Consider removing the cache step if it is not needed.`
     );
   }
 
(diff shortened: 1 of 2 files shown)


priya-kinthali (approved):

priyagupta108 (approved):

invisiblepancake (commented):
Change: Updated the error message to include a suggestion about no dependencies to cache.
Patch:
diff
- throw new Error(`Cache folder path is retrieved for ${packageManager} but doesn't exist on disk: ${cachePaths.join(', ')}`);
+ throw new Error(`Cache folder path is retrieved for ${packageManager} but doesn't exist on disk: ${cachePaths.join(', ')}. This likely indicates that there are no dependencies to cache. Consider removing the cache step if it is not needed.`);
src/cache-save.ts

Change: Updated the error message to include a suggestion about no dependencies to cache.
Patch:
diff
- )}`
+ )}. This likely indicates that there are no dependencies to cache. Consider removing the cache step if it is not needed.`

Scutua (approved):

PR #1090, opened 2025-04-21
Delete downloaded archive after extracting it during install

darktohka:
**Description:**

During a Python / Pypy / GraalPy installation, after the downloaded archive is extracted, it is not deleted. This leads to the archive taking up disk space, and often mistakenly getting archived together with the Python version.

This pull request ensures the downloaded archives are deleted after they are extracted and no longer needed.

**Check list:**

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -97011,6 +97011,7 @@ const httpm = __importStar(__nccwpck_require__(4844));
 const exec = __importStar(__nccwpck_require__(5236));
 const fs_1 = __importDefault(__nccwpck_require__(9896));
 const utils_1 = __nccwpck_require__(1798);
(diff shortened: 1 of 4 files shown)


darktohka:
> Comments suppressed due to low confidence (1)
> 
> **src/install-python.ts:143**
> 
>     * [nitpick] Consider updating the log message to be consistent with the other installers (e.g., use 'Deleting downloaded archive...' as in install-pypy.ts and install-graalpy.ts).
> 
> 
> ```
> core.info('Delete downloaded archive');
> ```

Since the `install-python.ts` workflow follows this convention (`Download from {url}`, `Extract downloaded archive`, `Execute installation script`) I decided to keep using the same convention for that workflow.

PR #1147, opened 2025-07-03
Use home dir instead of assuming /Users/runner

madrob:
**Description:**
Use actual user home directory instead of assuming that macOS runners use the `runner` user.

**Check list:**


madrob:
According to https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#macos the shared library configuration requires this path anyway.

PR #1149, opened 2025-07-11
Add overwrite input to optionally allow copying over existing dependency files from non-root directories

aparnajyothi-y:
**Description:**
Adds an overwrite input to prevent unintended overwrites when a composite action defines a cache-dependency-path (e.g., subdir/requirements.txt) that matches a file in the user’s workspace. By default, the copy is skipped if the target file exists, unless overwrite: true is explicitly set. This avoids silently replacing user files with the same name and path structure.

**Related issue:**
Add link to the related issue.

**Check list:**
- [X ] Mark if documentation changes are required.
- [ X] Mark if tests were added or updated to cover the changes.

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -96883,10 +96883,9 @@ function isGraalPyVersion(versionSpec) {
 }
 function cacheDependencies(cache, pythonVersion) {
     return __awaiter(this, void 0, void 0, function* () {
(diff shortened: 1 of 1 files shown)


v-aparnajyothi-y:
Closing this PR as it’s not required. The fix will be incorporated directly from the toolkit.

PR #1182, opened 2025-08-21
Change missing cache directory error to warning 

aparnajyothi-y:
Description:
This PR updates the cache lookup logic to log a warning instead of throwing an error when a cache folder path is not found on disk.

Previously: action failed with Error: Cache folder path ... doesn't exist.

Now: action logs a warning via core.warning() and continues execution.

Motivation: In many cases, missing cache simply means there are no dependencies to cache, which should not fail the workflow.

This improves developer experience by preventing unnecessary workflow failures while still surfacing useful diagnostic information.

Related issue:
https://github.com/actions/setup-python/issues/436


priya-kinthali (approved):

simondeziel (approved):
Thanks, LGTM!

PR #1284, opened 2026-03-09
Add cache-write input for read-only cache mode

salmanmkc:
Right now if you use `cache: pip` (or pipenv/poetry) in a PR workflow, the action restores *and* saves the cache. There's no way to get read-only mode where you benefit from existing caches without writing back. This matters for cache poisoning — an untrusted PR could plant bad packages in the cache that later get picked up by pushes to main.

This adds a `cache-write` input (defaults to `true`, no breaking change). Set it to `false` to skip the post-step save.

**Usage:**
```yaml
- uses: actions/setup-python@v6
  with:
    python-version: "3.12"
    cache: pip
    cache-write: ${{ github.event_name != 'pull_request' }}
```

**What changed:**
- `action.yml` — new `cache-write` input
- `src/cache-save.ts` — early return when `cache-write` is `false`
- `dist/` — rebuilt

Same change going into setup-node, setup-go, setup-java, setup-dotnet.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -20,6 +20,9 @@ inputs:
     default: ${{ github.server_url == 'https://github.com' && github.token || '' }}
   cache-dependency-path:
     description: "Used to specify the path to dependency files. Supports wildcards or a list of file names for caching multiple dependencies."
+  cache-write:
+    description: "Whether to save the cache at the end of the workflow. Set to false for cache read-only mode, useful for preventing cache poisoning from untrusted PR builds."
+    default: true
   update-environment:
     description: "Set this option if you want the action to update environment variables."
     default: true
--- a/dist/cache-save/index.js
+++ b/dist/cache-save/index.js
@@ -44170,6 +44170,11 @@ const cache_distributor_1 = __nccwpck_require__(92326);
 // https://github.com/actions/cache/pull/1217
 async function run(earlyExit) {
     try {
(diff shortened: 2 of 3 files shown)


brunoborges:
After reviewing the current GitHub Actions cache security model, I recommend closing this PR as unnecessary for the stated cache-poisoning threat.

- `pull_request` caches are scoped to `refs/pull/.../merge` and cannot be restored by the base branch or other pull requests.
- Low-trust events resolving to the default branch receive read-only cache access from the backend.
- Current `@actions/cache` releases honor backend cache modes and handle denied writes non-fatally.

The suggested `cache-write: ${{ github.event_name != 'pull_request' }}` pattern disables useful PR-local caching without adding protection for the default branch, and it remains true for `pull_request_target`, where backend enforcement is what provides protection.

Explicit restore-only behavior may still be useful ergonomically, but GitHub already recommends composing `actions/cache/restore` for that case rather than adding a security-specific setup-action input.

References:
- https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache
- https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-access-for-low-trust-workflow-triggers
- https://github.com/actions/toolkit/blob/main/packages/cache/RELEASES.md

